Issue #115, opened 2024-04-16
Including model config in initialization call for LanguageModel causes errors when calling model.trace


arunasank:
```python
## Initialization
    model_config = AutoConfig.from_pretrained(MODEL_PATH)
    tokenizer = AutoTokenizer.from_pretrained(MODEL_PATH, config=model_config, device_map="auto", padding_side="left")
    
    tokenizer.pad_token = tokenizer.eos_token
    model = LanguageModel(MODEL_PATH, config=model_config, device_map='auto') # Load the model

##nnsight call
 for prompt in prompts:
    prompt = tokenizer.decode(tokenizer(prompt, padding='max_length', max_length=MAX_LEN)["input_ids"])
    with model.trace(prompt, scan=False, validate=False) as tracer:
        for idx, row in mlp_ablate.iterrows():
            model.model.layers[row['layer']].mlp.down_proj.output[:, row['neuron']] = torch.tensor(row['values'])
        for idx, row in attn_ablate.iterrows():
            model.model.layers[row['layer']].self_attn.o_proj.output[:, row['neuron']] = torch.tensor(row['values'])
    token_ids = torch.topk(model.lm_head.output, 2, dim=-1).save()
    answers.append(tokenizer.batch_decode(token_ids))
```
cc @JadenFiotto-Kaufman

arunasank:
Error trace

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[26], line 46
     44 for prompt in prompts:
     45     prompt = tokenizer.decode(tokenizer(prompt, padding='max_length', max_length=MAX_LEN)["input_ids"])
---> 46     with model.trace(prompt, scan=False, validate=False) as tracer:
     47         for idx, row in mlp_ablate.iterrows():
     48             model.model.layers[row['layer']].mlp.down_proj.output[:, row['neuron']] = torch.tensor(row['values'])

File /mnt/align4_drive/arunas/miniconda3/envs/broca/lib/python3.12/site-packages/nnsight/contexts/Runner.py:49, in Runner.__exit__(self, exc_type, exc_val, exc_tb)
     47     self._graph = None
     48 else:
---> 49     super().__exit__(exc_type, exc_val, exc_tb)

File /mnt/align4_drive/arunas/miniconda3/envs/broca/lib/python3.12/site-packages/nnsight/contexts/Tracer.py:67, in Tracer.__exit__(self, exc_type, exc_val, exc_tb)
[...]

PR #117, opened 2024-04-19
Added kwargs to the tokenizer initialization

### LINKED ISSUE #116: kwargs are not passed to the tokenizer when initializing LanguageModel
```py
nn_model = LanguageModel("meta-llama/Llama-2-70b-hf", token=token)
```
Fails because `token` is not passed to the tokenizer initialization

Butanium:
Fixes #116

### DIFF
--- a/src/nnsight/models/LanguageModel.py
+++ b/src/nnsight/models/LanguageModel.py
@@ -151,7 +151,7 @@ class LanguageModel(GenerationMixin, NNsight):
         if self.tokenizer is None:
 
             self.tokenizer = AutoTokenizer.from_pretrained(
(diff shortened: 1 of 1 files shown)


JadenFiotto-Kaufman:
@Butanium I'm worried that overloading kwargs for the tokenizer and the model means an error will be thrown when either one of them dosent have the same kwarg? Maybe there should be a `tokenizer_args` keyword argument?

Butanium:
@JadenFiotto-Kaufman added it, let me know if I should change something

Butanium:
This PR now also fixes https://github.com/ndif-team/nnsight/issues/115

PR #144, opened 2024-05-28
add separate logging for remote execution, resolves #127

### LINKED ISSUE #127: Separate logging for remote execution
Right now remote execution status is printed out via print(). This should be logged instead. It should probably have a separate logger in logging.py that has a separate flag that can be configured in the config.

francescortu:
Add a separate logger for remote execution. Now it will log on stdout (replacing the print() statement)


JadenFiotto-Kaufman:
@francescortu Nice thank you! also just in the future make the PR to the dev branch.

PR #165, opened 2024-06-30
Fix missing parameter `names`

Xmaster6y:
## What does this PR do?

Fix missing parameter `names` in the `named_parameters` method.


src/nnsight/envoy.py
Xmaster6y, line 1: ## Additionally

In order to have a similar output also change this line:
https://github.com/ndif-team/nnsight/blob/107aab51f5473e73bb46029ae2e52660bdf331b5/src/nnsight/envoy.py#L238
into:
```python
envoys.append((self._module_path.lstrip("."), self))
```

PR #188, opened 2024-08-06
Unique Bridge Nodes

AdamBelfki3:
Adding functionality for optimizing bridged Intervention Graphs.

### Motivation

Previously, a `BridgeProtocol` node was created at each reference of an external proxy within the same graph. But optimally, all operations on the same proxy should all be linked to a single corresponding `BridgeProtocol` node.

#### Example

```python
from collections import OrderedDict
from nnsight import NNsight
import torch

input_size = 5
hidden_dims = 10
output_size = 2

net = torch.nn.Sequential(
    OrderedDict(
        [
            ("layer1", torch.nn.Linear(input_size, hidden_dims)),
            ("layer2", torch.nn.Linear(hidden_dims, output_size)),
        ]
    )
).requires_grad_(False)

input = torch.rand((1, input_size))

with model.session() as session:
    with model.trace(input) as tracer:
        l1_out = model.layer1.output

    with model.trace(input) as tracer_2:
        l1_out_double = l1_out * 2
        
        l1_out_sum = torch.sum(l1_out)

        tracer_2.graph.vis("Tracer_2")
```

Tracer_2

### Implementation

The `Bridge` structure tracks the bridge nodes created for a certain proxy, for every context it was externally referenced from.

Preprocessing of a `Node`'s arguments during instantiation no longer automatically creates a BridgeProtocol node in the case of an external reference to an argument. The `BridgeProtocol` handles directly bridging a node (creating a local reference) by checking first if the external node has been bridged previously before creating a new `BridgeProtocol` node.

When the `UpdateProtocol` is called on an external proxy node, the value of its local `BridgeProtocol` reference is also updated to reflect the new change.

### Results

Tracer_2

### DIFF
--- a/src/nnsight/tracing/Bridge.py
+++ b/src/nnsight/tracing/Bridge.py
@@ -1,9 +1,10 @@
 from collections import OrderedDict, defaultdict
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Dict, List, Optional
 
 from . import protocols
 
 if TYPE_CHECKING:
+    from ..intervention import InterventionProxy
     from .Graph import Graph
     from .Node import Node
 
@@ -15,7 +16,8 @@ class Bridge:
     Attributes:
         id_to_graph (Dict[int, Graph]): Mapping of graph id to Graph.
         graph_stack (List[Graph]): Stack of visited Intervention Graphs.
-        bridged_nodes (defaultdict[Node, List[Node]]): Mapping of bridged Nodes to the BridgeProtocol nodes representing them on different graphs. 
+        bridged_nodes (defaultdict[Node, defaultdict[int, Optional[InterventionProxy]]]): Mapping of bridged Nodes 
+            to the BridgeProtocol nodes representing them on different graphs. 
         locks (int): Count of how many entities are depending on ties between graphs not to be released.
     """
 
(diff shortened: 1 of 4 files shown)


src/nnsight/tracing/Node.py
```diff
@@ -105,7 +105,7 @@ def preprocess_node(node: Union[Node, Proxy]):
 
             if self.attached() and self.graph.id != node.graph.id:
 
-                node = protocols.BridgeProtocol.add(node).node
+                node = protocols.BridgeProtocol.bridge_node(node, self.graph.id)
```
JadenFiotto-Kaufman, line 108: Is bridge_node replacing add? If so why?

src/nnsight/tracing/Bridge.py
```diff
@@ -23,6 +25,7 @@ def __init__(self) -> None:
         self.id_to_graph: Dict[int, "Graph"] = OrderedDict()
         # Stack to keep track of most inner current graph
         self.graph_stack: List["Graph"] = list()
+        self.bridged_nodes: defaultdict[Node, List[Node]] = defaultdict(lambda: list())
```
JadenFiotto-Kaufman, line 28: Can you confirm there isnt any issue with this holding onto Nodes and therefore memory longer than needed?

src/nnsight/tracing/Node.py
AdamBelfki3: Yes it is. Because bridging a node should not necessarily involve adding a new `BridgeProtocol` node.  We would like it to use an existing one if applicable (to maintain the  invariant of one unique Bridge node per external proxy).

`.bridge_node()` calls `.add()` if a `BridgeProtocol` for the corresponding proxy does not exist yet.

src/nnsight/tracing/Bridge.py
AdamBelfki3: Are you saying we could reinitialize this storing variable every time a new intervention graph is being entered?

PR #189, opened 2024-08-10
Early Exit of Sequential Graph Based Contexts

AdamBelfki3:
New feature! You can now terminate early the execution of sequential graph-based contexts. The contexts concerned by this feature include the `Session` and `Iterator` contexts,`Tracer`context does not support it. Call `GraphBasedContext.exit()` to skip the remaining execution.

## Examples

Terminating a Session:

```python
with model.session() as session:
    l = session.apply(list).save()

    l.append(0)
    l.append(1)
    session.exit()
    l.append(2)

print("Result: ", l)
```
```python
"Result: [0, 1]"
```

Terminating an Iterator loop:

```python
with model.session() as session:
    l = session.apply(list).save()

    with session.iter([0, 1, 2, 3]) as (item, iterator):
        with item == 2:
            iterator.exit()
        
        l.append(item)

print("Result: ", l)
```
```python
"Result: [0, 1]"
```

## Implementation

The Exit operation makes use of the `EarlyStopProtocol`, creating a node the moment it is called. The `EarlyStopException` is caught when the `EarlyStopProtocol` node is executed, with the `Iterator` breaking out of its loop.

## Notes

An improvement should be put in place for the graph visualization feature to reflect sequential positioning of nodes in the graph.

### DIFF
--- a/src/nnsight/contexts/GraphBasedContext.py
+++ b/src/nnsight/contexts/GraphBasedContext.py
@@ -50,6 +50,15 @@ class GraphBasedContext(AbstractContextManager, BridgeMixin):
             kwargs=kwargs,
         )
     
+    def exit(self) -> InterventionProxy:
+        """ Exits the execution of a sequential intervention graph.
+        
+        Returns:
+            InterventionProxy: Proxy of the EarlyStopProtocol node.
+        """
+
+        return protocols.EarlyStopProtocol.add(self.graph)
+
     def vis(self, **kwargs) -> None:
         """
         Helper method to save a visualization of the current state of the intervention graph.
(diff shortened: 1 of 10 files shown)


src/nnsight/contexts/GraphBasedContext.py
```diff
@@ -68,17 +80,18 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:
...
-        graph.alive = False
+    def local_backend_execute(self) -> None:
 
-        if not isinstance(graph, weakref.ProxyType):
-            self.graph = weakref.proxy(graph)
+        try:
+            self.graph.execute()
+        except protocols.EarlyStopProtocol.EarlyStopException as e:
+            raise e
+        finally:
```
JadenFiotto-Kaufman, line 89: Does finally actually get called here? Because you are raising a new exception in except.

src/nnsight/contexts/session/Session.py
```diff
@@ -55,7 +56,10 @@ def iter(self, iterable, **kwargs) -> Iterator:
 
     def local_backend_execute(self) -> Dict[int, Graph]:
 
-        super().local_backend_execute()
+        try:
```
JadenFiotto-Kaufman, line 59: Why do you need to check for the Exception in all the subclasses if you are doing it in GraphBasedContext?

AdamBelfki3: Because in the case of the `Iterator`, we want to break out of the loop, which is done upon catching the exception.

src/nnsight/contexts/GraphBasedContext.py
AdamBelfki3: Yes, it's being called!

JadenFiotto-Kaufman (approved):

Issue #138, opened 2024-05-21
[Feature Request] Make token indexing work for batched input and UnifiedTransformer


Butanium:
Using `.token[0]` returns the padding token
```py
from nnsight import LanguageModel
model = LanguageModel("gpt2", device_map="cpu")

probs = model.trace("a zd zdb", trace=False).logits
with model.trace(["ab dfez zd", "a", "b"]):
    inp = model.input.save()
    inp2 = model.input[1]['input_ids'].t[0].save()
print(inp, inp2)
```
out:
```
(((), {'input_ids': tensor([[  397,   288,  5036,    89,  1976,    67],
         [50256, 50256, 50256, 50256, 50256,    64],
         [50256, 50256, 50256, 50256, 50256,    65]]), 'attention_mask': tensor([[1, 1, 1, 1, 1, 1],
         [0, 0, 0, 0, 0, 1],
         [0, 0, 0, 0, 0, 1]])}),
 tensor([  397, 50256, 50256]))
```

```
Name: nnsight
Version: 0.2.16
```

Butanium:
Oh wait, seems like token indexing is supposed to work only with `tracer.invoke` calls.
It would be nice if it also works on directly batched input, not sure how easy it is to add it with the current implementation

Butanium:
Ok but with UnifiedTransformer, token indexing doesn't work as padding side is right by default : 
```py
l = ["ab dfez zd", "a", "b"]
from nnsight import LanguageModel
model = LanguageModel("gpt2", device_map="cpu")

with model.trace(l):
    inp = model.input[1]['input_ids'].token[0].save()

with model.trace() as tracer:
    inp_l = []
    for s in l:
        with tracer.invoke(s):
            inp_l.append(model.input[1]['input_ids'].token[0].save())

from nnsight.models.UnifiedTransformer import UnifiedTransformer
umodel = UnifiedTransformer("gpt2", device="cpu")

with umodel.trace() as tracer:
    inp_l2 = []
    for s in l:
        with tracer.invoke(s):
            inp_l2.append(umodel.input[1]['input'].token[0].save())
print(inp) 
print([i.item() for i in inp_l])
print([i.item() for i in inp_l2])
```

Butanium:
@JadenFiotto-Kaufman I guess we can close this as UnifiedTransformer is no longer supported?

PR #203, opened 2024-08-22
Fix tokenizer kwarg in unifiedTransformer

### LINKED ISSUE #202: Cannot specify custom tokenizer with UnifiedTransformer
```py
!pip install -q git+https://github.com/ndif-team/nnsight@0.3 transformer_lens
from nnsight.models.UnifiedTransformer import UnifiedTransformer
from transformers import AutoTokenizer
tokenizer = AutoTokenizer.from_pretrained("gpt2")
model = UnifiedTransformer("gpt2", tokenizer=tokenizer, processing=False)


---> 43         super().__init__(hooked_model, tokenizer=self.tokenizer, *args, **kwargs)
     44 
     45     def _prepare_inputs(

TypeError: nnsight.models.LanguageModel.LanguageModel.__init__() got multiple values for keyword argument 'tokenizer'
```

Solution (PR incoming): check if tokenizer is in kwargs.

Butanium:
fixes #202

### DIFF
--- a/src/nnsight/models/UnifiedTransformer.py
+++ b/src/nnsight/models/UnifiedTransformer.py
@@ -4,7 +4,7 @@ from typing import Any, Dict, List, Tuple, Union
 
 import torch
 from transformer_lens import HookedTransformer, HookedTransformerConfig
-from transformers import BatchEncoding
+from transformers import BatchEncoding, PreTrainedTokenizer
 
 from .LanguageModel import LanguageModel
 
@@ -28,13 +28,13 @@ class UnifiedTransformer(LanguageModel):
     """
 
     def __init__(
-        self, model: str, *args, processing: bool = True, **kwargs
+        self, model: str, *args, processing: bool = True, tokenizer: Optional[PreTrainedTokenizer] = None, **kwargs
     ) -> None:
         if processing:
-            hooked_model = HookedTransformer.from_pretrained(model, *args, **kwargs)
+            hooked_model = HookedTransformer.from_pretrained(model, *args, tokenizer=tokenizer, **kwargs)
         else:
             hooked_model = HookedTransformer.from_pretrained_no_processing(
-                model, *args, **kwargs
+                model, *args, tokenizer=tokenizer, **kwargs
             )
 
         self.tokenizer = hooked_model.tokenizer
--- a/src/nnsight/models/UnifiedTransformer.py
+++ b/src/nnsight/models/UnifiedTransformer.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Dict, List, Tuple, Union
+from typing import Any, Dict, List, Tuple, Union, Optional
 
 import torch
 from transformer_lens import HookedTransformer, HookedTransformerConfig


src/nnsight/models/UnifiedTransformer.py
```diff
@@ -28,13 +28,13 @@ class UnifiedTransformer(LanguageModel):
...
+        self, model: str, *args, processing: bool = True, tokenizer: Optional[PreTrainedTokenizer] = None, **kwargs
     ) -> None:
         if processing:
-            hooked_model = HookedTransformer.from_pretrained(model, *args, **kwargs)
+            hooked_model = HookedTransformer.from_pretrained(model, *args, tokenizer=tokenizer, **kwargs)
         else:
             hooked_model = HookedTransformer.from_pretrained_no_processing(
-                model, *args, **kwargs
+                model, *args, tokenizer=tokenizer, **kwargs
             )
```
Butanium, line 39: we could also 
```py
kwargs.pop("tokenizer", None)
```
here if you prefer

src/nnsight/models/UnifiedTransformer.py
```diff
@@ -28,13 +28,13 @@ class UnifiedTransformer(LanguageModel):
...
 
         self.tokenizer = hooked_model.tokenizer
```
Butanium, line 40: also why do you do that when it'll be done again in the LanguageModel init?

Butanium:
Might be a good time to also fix https://github.com/ndif-team/nnsight/issues/138 by enforcing `padding_side="left"` for unified_transformer too ?

src/nnsight/models/UnifiedTransformer.py
JadenFiotto-Kaufman: Sorry what is the context behind this?

src/nnsight/models/UnifiedTransformer.py
```diff
@@ -28,13 +28,13 @@ class UnifiedTransformer(LanguageModel):
     """
 
     def __init__(
-        self, model: str, *args, processing: bool = True, **kwargs
+        self, model: str, *args, processing: bool = True, tokenizer: Optional[PreTrainedTokenizer] = None, **kwargs
```
JadenFiotto-Kaufman, line 31: You need to import Optional. After that I'll merge it in!

Butanium: done!

src/nnsight/models/UnifiedTransformer.py
Butanium: in unifiedtransformer init, you set self.tokenizer and then call the super init which will run https://github.com/ndif-team/nnsight/blob/f64d8a963d72acd58f2eaae6b1a2fcbf4d29cb63/src/nnsight/models/LanguageModel.py#L141
which will set self.tokenizer again, right?

JadenFiotto-Kaufman:
@Butanium Should we just get rid of enforcing padding side all together?

Butanium:
@JadenFiotto-Kaufman I think we can probably do it but we'd need to ensure that the token indexing actually work in both cases

Butanium:
@JadenFiotto-Kaufman could we merge this as is ? We can still take a decision regarding padding side later but it'd be nice to have this fixed asap

JadenFiotto-Kaufman:
@Butanium 👍

PR #229, opened 2024-09-03
Default Tokenizer Padding Side

AdamBelfki3:


src/nnsight/models/LanguageModel.py
```diff
@@ -175,6 +175,9 @@ def _load(
             if tokenizer_kwargs is None:
                 tokenizer_kwargs = {}
 
+            if "padding_side" not in tokenizer_kwargs:
+                tokenizer_kwargs["padding_side"] = "left"
+
```
Butanium, line 180: maybe add a warning if the set kwarg is not left explaining why it could go wrong?

PR #250, opened 2024-09-27
Streaming protocol

JadenFiotto-Kaufman:
My crowning achievement.

### DIFF
--- a/src/nnsight/__init__.py
+++ b/src/nnsight/__init__.py
@@ -548,3 +548,4 @@ dict = GlobalTracingContext.GLOBAL_TRACING_CONTEXT.dict
 apply = GlobalTracingContext.GLOBAL_TRACING_CONTEXT.apply
 log = GlobalTracingContext.GLOBAL_TRACING_CONTEXT.log
 cond = GlobalTracingContext.GLOBAL_TRACING_CONTEXT.cond
+upload = GlobalTracingContext.GLOBAL_TRACING_CONTEXT.upload
--- a/src/nnsight/__init__.py
+++ b/src/nnsight/__init__.py
@@ -1,4 +1,4 @@
-# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # 
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 #                                                                                                                       #
 #      ::::    ::: ::::    :::  :::::::: ::::::::::: ::::::::  :::    ::: :::::::::::       :::::::       ::::::::      #
 #      :+:+:   :+: :+:+:   :+: :+:    :+:    :+:    :+:    :+: :+:    :+:     :+:          :+:   :+:     :+:    :+:     #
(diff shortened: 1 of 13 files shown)


src/nnsight/contexts/Tracer.py
```diff
@@ -179,6 +180,9 @@ def remote_backend_handle_result_value(self, value: Dict[str, Any]) -> None:
         # TODO : graph mismatch handle. hash json ?
         for node_name, node_value in value.items():
             self.graph.nodes[node_name]._value = node_value
+            
+    def remote_backend_get_stream_node(self, name: str, graph_id: str) -> "Node":
```
MichaelRipa, line 184: I get that the function name provides context for when the function is called (and perhaps the only actual usecase), but why not use a more general name, like `get_node()`? It is just a getter function afterall

src/nnsight/contexts/backends/RemoteBackend.py
```diff
@@ -80,28 +107,36 @@ def __init__(
...
+        model_key = self.object.remote_backend_get_model_key()
 
         from ...schema.Request import RequestModel
 
         # Create request using pydantic to parse the object itself.
-        return RequestModel(object=obj, model_key=model_key)
+        return RequestModel(object=self.object, model_key=model_key)
 
-    def __call__(self, obj: RemoteMixin):
+    def __call__(self, object: RemoteMixin):
```
MichaelRipa, line 131: This function is way easier to read now!

src/nnsight/contexts/backends/RemoteBackend.py
```diff
@@ -110,25 +145,27 @@ def __call__(self, obj: RemoteMixin):
...
+            self.non_blocking_request(request)
 
-            self.non_blocking_request(request=request)
+        # Cleanup
+        self.object.remote_backend_cleanup()
 
-        obj.remote_backend_cleanup()
-
-    def handle_response(self, data: Any) -> "ResponseModel":
+    def handle_response(self, response: "ResponseModel") -> None:
```
MichaelRipa, line 159: For readability, I think that `blocking_request()` and `non_blocking_request()` should be placed immediately after `__call__()`

src/nnsight/__init__.py
```diff
@@ -548,3 +549,88 @@ def set_module_tensor_to_device(
...
+    Returns:
+        Callable: Traceable function.
+    """
+
+    @wraps(fn)
+    def inner(*args, **kwargs):
+
+        return apply(fn, *args, **kwargs)
+
+    return inner
```
MichaelRipa, line 575: I see this pattern defined in numerous locations (`__init__.py`, `GraphBasedContext.py`). Do you think this could be reused as a `util` function (with a descriptive docstring)?

src/nnsight/schema/format/types.py
```diff
@@ -110,13 +116,14 @@ def deserialize(self, handler: DeserializeHandler) -> Node:
             target=self.target.deserialize(handler),
             args=[try_deserialize(value, handler) for value in self.args],
             kwargs={
-                key: try_deserialize(value, handler) for key, value in self.kwargs.items()
+                key: try_deserialize(value, handler)
```
MichaelRipa, line 119: I know it is slightly less efficient, but I think we should avoid adding formatting changes that don't effect functionality to PR's which are doing things changing functionality. Especially in scripts not containing any functional changes.

In an optimal setup, we'd already have a consistent formatting style, and each PR would have a check for linting. In our current circumstances, use `git add -patch {file}` to go through and first add functional changes, to push to one brach. Then commit the rest to a seperate branch and make a seperate PR. Or just routinely spend some time going through the code and making little cleanups and push

MichaelRipa: wait nvm I see there is a flake8 linter in the checks

PR #257, opened 2024-10-04
added missing import of annotations

Butanium:


Butanium:
https://github.com/ndif-team/nnsight/pull/259 is actually the fix

PR #261, opened 2024-10-08
Update Node.py

loftusa:
I'm going to make small PRs whenever I find the error message confusing. Let me know if you would like me to change the wording on this.


JadenFiotto-Kaufman:
@loftusa This exception catching is not primarily hit what not accessing .output or .input. Also @AdamBelfki3 has totally refactored the exception catching in the next update!

PR #286, opened 2024-11-10
interleaving: support no nodes in graph

ainatersol:


JadenFiotto-Kaufman:
@ainatersol Good catch thank you!

PR #288, opened 2024-11-10
set up stop functionality

ainatersol:


JadenFiotto-Kaufman:
@ainatersol  This is great thank you! I don't think we want stop on the Envoy class but everything else looks great. I'd probably also break in the Graph.clean loop if we encounter an already executed Node https://github.com/ndif-team/nnsight/pull/288/files#diff-b236b26f299497377cba19c2e94031f29174e4e1b399f4bd81d8cb880bfaf1c5R70

I can handle adding these changes.

PR #306, opened 2024-12-08
Handle Frame Injection in Inline Trace

AdamBelfki3:


JadenFiotto-Kaufman:
@AdamBelfki3 https://github.com/ndif-team/nnsight/pull/309/files

PR #307, opened 2024-12-13
Doc fixes 20241213

BorisTheBrave:


pyproject.toml
```diff
@@ -11,7 +11,7 @@ authors = [
 ]
 description = "Package for interpreting and manipulating the internals of deep learning models."
 readme = "README.md"
-requires-python = ">=3.7"
+requires-python = ">=3.10"
```
BorisTheBrave, line 14: The docs wouldn't build with 3.9 as you have typehints that use `a | b` which was only introduced in 3.10.

JadenFiotto-Kaufman:
@BorisTheBrave Thank you!

PR #309, opened 2024-12-14
fix tracing=False no longer working

### LINKED ISSUE #308: `trace=False`  is broken in 0.4
```py
model = LanguageModel("gpt2")
out = model.trace("hello", trace=False)
```
fails with
```py
Cell In[5], line 6
      2 from nnsight import LanguageModel
      4 model = LanguageModel("gpt2")
----> 6 out = model.trace("hello", trace=False)

nnsight/modeling/mixins/remoteable.py:38, in RemoteableMixin.trace(self, method, backend, remote, blocking, trace, scan, *inputs, **kwargs)
     34 elif isinstance(backend, str):
     35     backend = RemoteBackend(
     36         self.to_model_key(), host=backend, blocking=blocking
     37     )
---> 38 return super().trace(
     39     *inputs,
     40     method=method,
     41     backend=backend,
     42     trace=trace,
     43     scan=scan,
     44     **kwargs,
     45 )

nnsight/intervention/base.py:202, in NNsight.trace(self, trace, scan, method, invoker_kwargs, backend, *inputs, **kwargs)
    198         with tracer:
    200             output = self._envoy.output.save()
--> 202         return output.value
    204 # If trace is False, you had to have provided an input.
    205 if not trace:

AttributeError: 'CausalLMOutputWithCrossAttentions' object has no attribute 'value'
```
probably worth adding a test for `trace=False`

Butanium:
fixes #308

### DIFF
--- a/src/nnsight/intervention/base.py
+++ b/src/nnsight/intervention/base.py
@@ -11,7 +11,8 @@ from .backends import NoopBackend
 from .contexts import EditingTracer, InterleavingTracer, Session
 from .envoy import Envoy
 from .graph import (InterventionGraph, InterventionNode, InterventionProxy,
-                    InterventionProxyType, Proxy)
+                    InterventionProxyType)
+from .graph.proxy import Proxy
 from .interleaver import Interleaver
 from .. import CONFIG
 
--- a/src/nnsight/intervention/base.py
+++ b/src/nnsight/intervention/base.py
@@ -199,7 +199,7 @@ class NNsight:
 
                     output = self._envoy.output.save()
 
-                return output.value
+                return output
 
         # If trace is False, you had to have provided an input.
         if not trace:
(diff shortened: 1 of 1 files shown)


Butanium:
Btw no idea what you're doing with this custom backend, but this test https://github.com/ndif-team/nnsight/blob/7664e72b4a0b6d89098f4fec50365a185e130f19/tests/test_tiny.py#L43 won't work in 0.4 as it uses `.value`.
Therefore I'm unsure how to write a test for this `tracing=False` behavior

JadenFiotto-Kaufman:
@Butanium  Can you just add a check like:

```py

if isinstance(output, Proxy):
    output = output.value
```

Theres an option to turn off the automatic value injection (or with different backends), so if thats off id like this to still return the real value. After that ill merge it in thanks!!

Butanium:
done!

PR #315, opened 2025-01-27
fixed super() call in `BridgeException`

jinensetpal:
`super` isn't actually called in `BridgeException`, the proposed change is a minor syntax correction to fix that.


JadenFiotto-Kaufman:
@jinensetpal Bridge is no longer a thing in 0.4. Thank you though!

PR #373, opened 2025-03-18
Bug fix: ModuleList input and outputs

MichaelRipa:
## Description:

`Envoy` had a small bug where modules of type `ModuleList` would have their inputs and outputs returned without first being appended to their respective stacks.


JadenFiotto-Kaufman:
@MichaelRipa  Going to remake this PR to dev. Also check out my changes

Issue #382, opened 2025-03-24
Bug: 0-dim arguments


MichaelRipa:
## Description

Suppose you have the following model:

```
import torch
import torch.nn as nn

class ZeroDModel(nn.Module):
    """A simple model which adds a small random tensor to the input."""
    def __init__(self, epsilon: float = 0.01):
        super().__init__()
        self.epsilon = epsilon
        
    def forward(self, input: torch.Tensor):

        hidden = torch.randn(input.shape, dtype=torch.float32)
        return (self.epsilon * hidden) + input
```

It just so happens to be implemented in such a way that it can take a 0d tensor as a valid input to its `forward()` method.

Although trivial, this model should still work with the `NNsight` base class. However, given the following script:

```
import nnsight

model = nnsight.NNsight(ZeroDModel())

# This should work
input_tensor = torch.tensor(1, dtype=torch.float32)

with model.trace(input_tensor):
    output = model.output.save()
```

It results in the following error:

```
    self.inputs, self.batch_size = self.tracer._model._prepare_input(
  File ".../nnsight/src/nnsight/intervention/base.py", line 468, in _prepare_input
    return (args, kwargs), len(args[0])
  File ".../python3.10/site-packages/torch/_tensor.py", line 1132, in __len__
    raise TypeError("len() of a 0-d tensor")
TypeError: len() of a 0-d tensor
```

JadenFiotto-Kaufman:
Interesting.

PR #383, opened 2025-03-24
Added tests for models which are kwargs only

MichaelRipa:
## Description

This PR introduces a new set of tests for models which only take keyword arguments (enforced via "*" in their `forward()` method).

**Note:** In creating these tests, I stumbled upon a bug, which I raised an issue for (#382 ). I added a `skipif` to one of the tests which should be removed when the bug is patched.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -28,4 +28,4 @@ def load_pyproject_toml():
     except toml.TomlDecodeError as e:
         pytest.fail(f"Failed to load pyproject.toml: {e}")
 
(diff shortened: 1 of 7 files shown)


JadenFiotto-Kaufman:
@MichaelRipa  Can you make this PR into dev instead of main and ill merge it?

JadenFiotto-Kaufman:
@MichaelRipa Please make a new PR off dev with just the test changes

PR #384, opened 2025-03-24
Better error messages for accessing outside of trace context

MichaelRipa:
## Description

This PR introduces more informative error messages when trying to access `.input`, `.inputs` or `.output` outside of a trace context.

### DIFF
--- a/tests/test_lm.py
+++ b/tests/test_lm.py
@@ -491,3 +491,17 @@ def test_undispatched_extra_module(device: str, MSG_prompt: str):
         output = model.generator.output.save()
 
     output.value
+
+
+def test_envoy_input_output_access(gpt2: nnsight.LanguageModel, MSG_prompt: str):
+    with gpt2.trace(MSG_prompt) as tracer:
+        l1_out = gpt2.transformer.h[0].output.save()
+
+    with pytest.raises(RuntimeError):
+        gpt2.transformer.h[0].output
+
+    with pytest.raises(RuntimeError):
+        gpt2.transformer.h[0].input
+
+    with pytest.raises(RuntimeError):
+        gpt2.transformer.h[0].inputs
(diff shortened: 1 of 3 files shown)


AdamBelfki3:
Did you run into limitations when you tried to decorate the functions in question? Is it because they're already decorated as properties?

tests/test_tiny.py
```diff
@@ -314,3 +314,17 @@ def test_torch_creation_operations_patch(tiny_model: NNsight, tiny_input: torch.
         torch.randn(l1_output.shape)
         torch.randperm(l1_output.shape[0])
         torch.zeros(l1_output.shape)
+
+
+def test_envoy_input_output_access(tiny_model: NNsight, tiny_input: torch.Tensor):
+    with tiny_model.trace(tiny_input) as tracer:
```
AdamBelfki3, line 320: A previous tracing call should not be necessary to test this behavior

tests/test_lm.py
```diff
@@ -491,3 +491,17 @@ def test_undispatched_extra_module(device: str, MSG_prompt: str):
         output = model.generator.output.save()
 
     output.value
+
+
+def test_envoy_input_output_access(gpt2: nnsight.LanguageModel, MSG_prompt: str):
+    with gpt2.trace(MSG_prompt) as tracer:
+        l1_out = gpt2.transformer.h[0].output.save()
```
AdamBelfki3, line 498: A previous tracing call should not be necessary to test this behavior

AdamBelfki3 (changes requested):

JadenFiotto-Kaufman:
@MichaelRipa  Can you also make this error message better in this PR? https://github.com/ndif-team/nnsight/blob/149e14c12dde9267868051878e7091011b27fe62/src/nnsight/tracing/graph/node.py#L143

MichaelRipa:
@JadenFiotto-Kaufman The node error message shows up in undesirable circumstances:

```
import nnsight
model = nnsight.LanguageModel("openai-community/gpt2")

with model.trace("ayy"):
  x = model.transformer.h[0].attn.attn_dropout.output.save()
    
print(x.shape)
```

```
ValueError: Accessing value before it's been set. Did you forget to call .save()?
```

Is this expected behavior (and if so would there be a better error to log) or a bug?

PR #417, opened 2025-04-22
Add image to image diffusion

Kmoneal:
This is very similar to Diffusion but instead of `seed` takes `image` of the types specified by the model. For Stable Diffusion, accepted types can be found [here](https://huggingface.co/docs/diffusers/v0.17.1/api/pipelines/stable_diffusion/img2img#diffusers.StableDiffusionImg2ImgPipeline.__call__.image).

I'm happy to use this to kick off conversations on this topic as well.

### DIFF
--- a/src/nnsight/modeling/img2img_diffusion.py
+++ b/src/nnsight/modeling/img2img_diffusion.py
@@ -1,0 +1,119 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
+
+import torch
+from diffusers import AutoPipelineForImage2Image
+from transformers import BatchEncoding
+from typing_extensions import Self
+from ..intervention.contexts import InterventionTracer
+
+from .. import util
+from .mixins import RemoteableMixin
+
+
+class Diffuser(util.WrapperModule):
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__()
+
+        self.pipeline = AutoPipelineForImage2Image.from_pretrained(*args, **kwargs)
+        
+        for key, value in self.pipeline.__dict__.items():
+            if isinstance(value, torch.nn.Module):
+                setattr(self, key, value)
+
+        self.tokenizer = self.pipeline.tokenizer
+
+
+class Img2ImgDiffusionModel(RemoteableMixin):
+    
+    __methods__ = {"generate": "_generate"}
+
+    def __init__(self, *args, **kwargs) -> None:
+
+        self._model: Diffuser = None
+
+        super().__init__(*args, **kwargs)
+        
+    def _load_meta(self, repo_id:str, **kwargs):
+        
+        
+        model = Diffuser(
+            repo_id,
+            device_map=None,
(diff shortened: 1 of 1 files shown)


ericluo04:
Hi @Kmoneal - thank you very much for creating this PR! 

Unfortunately, I've been having some trouble getting this implementation to work. Would it be kindly possible for you to share a minimally reproducible example of how to do the following (e.g., with `stabilityai/stable-diffusion-3.5-large`):
1. Store the activations of the residual stream (e.g., output of the transformer block at index 24), for any choice/range of timestep. 
2. Intervene on the activations of the above (e.g., by tripling the activation of a particular dimension), for any choice/range of timestep.

Any help would be much appreciated! Thanks again. :)

ericluo04:
Figured it out! Turns out you can't specify the prompt using a parameter but just have to enter it directly first. See below for extracting the residual stream of the 24th layer in `stabilityai/stable-diffusion-3.5-large`:

```
# transformer block layers
layers = pipe.transformer.transformer_blocks

with pipe.generate("", negative_prompt="", guidance_scale=7.5, 
                   image=init_image, width=832, height=1248,
                   strength=.5, num_inference_steps=4,
                   seed=None):
    # initialize list to store activations
    res_stream = nnsight.list().save() 
    res_stream_text = nnsight.list().save()
    res_stream_image = nnsight.list().save()
    
    # loop over steps
    with layers.iter[0:1]:
        # 24th layer output residual stream for text and image stream
        res_stream.append(layers[24].output)
        # 25th layer input for text stream (to check if same as above)
        res_stream_text.append(layers[25].norm1_context.input)
        # 25th layer input for image stream (to check if same as above)
        res_stream_image.append(layers[25].norm1.input)
```

JadenFiotto-Kaufman:
Hi @Kmoneal, and sorry this sat so long. You asked for a conversation and got
seventeen months of silence, which is our fault.

The need was real and it is now handled, though not the way this PR proposed. On
`0.8` there is one generic `DiffusionModel` that wraps any diffusers pipeline, so
a per-task subclass is the shape the rewrite moved away from. What was missing was
any way to tell it which pipeline class to load, since `_load` hardcoded
`DiffusionPipeline`, which builds whatever class the repo declares. For every
Stable Diffusion repo that is the text-to-image one, so `image=` had nowhere to
go.

#712 adds `automodel=`:

```python
from diffusers import AutoPipelineForImage2Image

model = DiffusionModel(repo, automodel=AutoPipelineForImage2Image)

with model.trace("a photo of a cat", image=source, strength=0.5,
                 num_inference_steps=20):
    latents = model.unet.output[0].save()
    result = model.output.save()
```

An `AutoPipelineFor*` picks the right class per architecture, so the same line
reaches image-to-image on Flux, SD3 and SDXL as well. Inpainting works through
`AutoPipelineForInpainting`, and a concrete class or its name in `diffusers` works
too if you want to be exact.

Closing this in favour of that. Thank you for raising it, and for the pointer to
the accepted image types in the original description, which is what made it clear
this was about the pipeline class rather than about the argument.

PR #437, opened 2025-06-16
0.5 Improvements + Tests

AdamBelfki3:
Improvements:
- recursive source of methods
- skipping modules with multiple invokers
- caching with multiple invokers
- fix skip if you call the same module's input after skipping
- iterating over token generation with index returned
- fix caching after a module is called

Tests:
- scanning
- out of order interventions
- source
- module skipping
- iterating over token generation
- activations cache
- renaming

### DIFF
--- a/src/nnsight/intervention/interleaver.py
+++ b/src/nnsight/intervention/interleaver.py
@@ -630,10 +630,19 @@ class Mediator:
             raise SkipException(value)
         
         else:
-            self.history.add(provider)
-            self.event_queue.put((Events.SKIP, (requester, value)))
-            
-            return False
+            if requester in self.history:
+                self.respond(
+                    ValueError(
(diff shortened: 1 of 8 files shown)


src/nnsight/intervention/interleaver.py
```diff
@@ -82,6 +82,8 @@ def __init__(self, invokers: List[Mediator], tracer: InterleavingTracer, batcher
 
         self.state = dict()
         self.iteration_tracker = defaultdict(int)
+
+        self.user_cache: Optional[Cache] = user_cache
```
JadenFiotto-Kaufman, line 86: So both the interleaver and mediators have their own user cache?

AdamBelfki3: yes, this is if you are tracing with multiple invokers and want to cache outside of the invokers to capture the entire batch. So the interleaver needs to have its own cache independent of the mediators.

PR #442, opened 2025-06-17
Cache Attribute Access

AdamBelfki3:
Saved activations in the cache object can now be accessed with their module path from the original model.

```py
with model.trace("_") as tracer:
    cache = tracer.cache()

assert torch.equal(cache.model.transformer.h[0].mlp.output, cache["model.transformer.h.0.mlp"].output)
```

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,6 +1,7 @@
 [pytest]
 markers =
     config: mark a test as part of configuration tests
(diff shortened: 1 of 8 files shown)


AdamBelfki3:
@JadenFiotto-Kaufman Will not be merged?

AdamBelfki3:
okay I see it nvm haha

PR #444, opened 2025-06-17
feature: access cache content as attributes

JadenFiotto-Kaufman:


src/nnsight/intervention/tracing/tracer.py
```diff
@@ -42,6 +42,62 @@ class Entry:
         def input(self):
             return [*self.inputs[0], *self.inputs[1].values()][0]
 
+    class CacheDict(Dict):
```
JadenFiotto-Kaufman, line 45: I probably wouldn't use a dict anymore for cache. Some object like CacehEntry that has attribute children. Then Define __getitem__ to allow string access.

AdamBelfki3: Might be useful for users to get a dictionary object to interact with? To be able to get all items (with key values) fast?

src/nnsight/intervention/tracing/tracer.py
```diff
@@ -42,6 +42,62 @@ class Entry:
...
+        """
+
+        def __init__(self, data: "Union[Cache.CacheDict, Dict[str, Cache.Entry]]", path: Optional[str] = ""):
+            self._path = path
+
+            super().__init__(data)
+
+        def __getitem__(self, key):
+            if isinstance(key, str):
+                return dict.__getitem__(self, key)
```
AdamBelfki3, line 77: Append path to this cache reference

PR #445, opened 2025-06-17
Source only accessible during tracing

AdamBelfki3:
The .source functionality is now only accessible during tracing.

```py
print(model.transformer.h[0].attn.source)

>>> ValueError
```

```py
with model.scan("_"):
    print(model.transformer.h[0].attn.source)
```


JadenFiotto-Kaufman:
@AdamBelfki3 Did we still want to do this?

PR #465, opened 2025-07-01
Handle source of decorated forward methods

AdamBelfki3:


JadenFiotto-Kaufman:
@AdamBelfki3 Does this keep the original wrap? Like can you confirm the decorator of the method is still on after our inject?

PR #513, opened 2025-09-03
Pass through Lora Request Kwargs

abhay-sheshadri:
Added support for passing through LORA Requests into vllm.trace

```python
# Test with LoRA adapter
with vllm.trace(test_prompts, temperature=0.0, max_tokens=500, lora_request=lora_request) as tracer:
    lora_results = vllm.generator.output.save()
```


JadenFiotto-Kaufman:
Should already work.

PR #523, opened 2025-09-17
Fix: normalize indentation before compile

rimon15:
Currently, if indentation within trace context is != 4, there is an error due to the way the code is concatenated and compiled (indentation error). It may be that several users (myself included), use, for example, 2 spaces for formatting. The system is currently incompatible with any indentation scheme that is not 4 spaces. This fix would normalize the code block using `autopep8`, ensuring that the execution is indentation-agnostic.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ dependencies = [
     "accelerate",
     "toml",
     "ipython",
+    "autopep8",
     ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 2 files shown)


JadenFiotto-Kaufman:
@rimon15 Can you checkout the latest version I just pushed `v0.5.5`? I think I fixed it there. 

This diff specifically: https://github.com/ndif-team/nnsight/commit/4f2ea38f183b69f802029d794d289551072a136e#diff-71613f69ec0586bb50b95607267f05e44bb6e557193178775bc287d5c244bb63L20-L216

rimon15:
Confirmed, this is working now in the latest version.

PR #552, opened 2025-11-13
Cache ignores not found alias without failing

### LINKED ISSUE #535: Unused renaming scheme make cache access via renamed name fail
Even if a renaming is not used (`model.language_model` here):
```py

m = LanguageModel(
    "yujiepan/llama-3.2-tiny-random",
    rename={"model": "foo", "model.language_model": "foo"},
    # rename={"model": "foo"},  # works
)
with m.trace("Hello, world!") as tracer:
    cache = tracer.cache(modules=[m.foo.layers[0]])
print(cache.model.foo.layers[0].output)
```
It will make caching fail:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[7], line 10
      8 with m.trace("Hello, world!") as tracer:
      9     cache = tracer.cache(modules=[m.foo.layers[0]])
---> 10 print(cache.model.foo.layers[0].output)

File ~/projects/nnterp/.venv/lib/python3.10/site-packages/nnsight/intervention/tracing/tracer.py:156, in Cache.CacheDict.__getattr__(self, attr)
    154     name = self._alias[attr]
    155     name = name.removeprefix(".")
--> 156     return self.__getattr__(name)
    157 else:
    158     raise AttributeError(f"'{attr}' module path was never cached. '{self.__class__.__name__}' has no matching attribute.")

File ~/projects/nnterp/.venv/lib/python3.10/site-packages/nnsight/intervention/tracing/tracer.py:158, in Cache.CacheDict.__getattr__(self, attr)
    156     return self.__getattr__(name)
    157 else:
--> 158     raise AttributeError(f"'{attr}' module path was never cached. '{self.__class__.__name__}' has no matching attribute.")

AttributeError: 'model.language_model' module path was never cached. 'CacheDict' has no matching attribute.
```

AdamBelfki3:


JadenFiotto-Kaufman:
Closing this as superseded. The behaviour you wanted is now the documented
default on `0.8`, reached by a different route.

The file this patches, `src/nnsight/intervention/tracing/tracer.py`, does not
exist on `0.8`, and neither does `tests/test_lm.py`. The `_alias` and
`alias_to_name` machinery it walks is gone too, replaced by `_rename` and
`_aliases` on the envoy.

On the behaviour: a rename key that resolves to nothing is skipped silently, and
that is deliberate rather than incidental. `_bind_aliases` does
`try: target = self.get(path)` / `except AttributeError: continue`. It is what
lets one rename dict cover architectures that spell a module differently, as in
`{"attn": "att", "self_attn": "att"}`, which binds whichever of the two exists.
That pattern came up again in #701 as the thing not to break when we started
raising on aliases that would shadow something.

Worth recording, since it is the fourth PR to circle this: #535 reported the
original cache-under-rename problem, #682 rewrote navigation to fix it, this one
fixed the unresolvable-key half, and #701 landed the shadowing checks today. Four
independent reports on one subsystem over ten months. It is in better shape now,
but if you have a view on whether the current alias resolution is covered end to
end rather than only at the unit level, that would be worth an issue.

PR #620, opened 2026-03-02
Fix stdlib modules being registered for pickle-by-value in get_local_env()

### LINKED ISSUE #619: Remote trace serialization fails with 'cannot pickle classmethod_descriptor' due to stdlib modules registered for pickle-by-value
## Description

Remote tracing fails with `TypeError: cannot pickle 'classmethod_descriptor' object` on nnsight 0.6.1. Even the most minimal remote trace reproduces the issue:

```python
from nnsight import CONFIG, LanguageModel

CONFIG.set_default_api_key("...")
model = LanguageModel("openai-community/gpt2")

with model.trace("Hello", remote=True):
    output = model.lm_head.output.save()
```

## Root Cause

The bug is in `get_local_env()` in `ndif.py` (lines ~392-411). It uses `sysconfig.get_paths()["stdlib"]` to filter out stdlib modules from the "local modules" list, but on some platforms (confirmed on macOS with Homebrew Python 3.12), the path returned by `sysconfig` doesn't match the paths returned by `pkgutil.iter_modules()`. This causes **all stdlib modules** to be misidentified as "local":

```python
# ndif.py get_local_env()
stdlib_path = sysconfig.get_paths()["stdlib"]
# ...
for importer, module_name, is_pkg in pkgutil.iter_modules():
    # ...
    if module_path.startswith(stdlib_path):  # <-- This check fails on some platforms
        continue
    packages[module_name] = "local"  # stdlib modules end up here
```

Then in `pull_env()`, all packages with version `"local"` get registered with cloudpickle's `_PICKLE_BY_VALUE_MODULES`:

```python
for package, version in local_env.get("packages", {}).items():
    if version == "local":
        register(package)  # adds to _PICKLE_BY_VALUE_MODULES
```

This causes `collections` (among others) to be registered for pickle-by-value. When cloudpickle then tries to serialize a `torch.Tensor` (whose `__reduce_ex__` returns an `OrderedDict`), it attempts to serialize `collections.OrderedDict` inline rather than by reference. This encounters `OrderedDict.fromkeys`, which is a `classmethod_descriptor` that cannot be pickled.

## Confirmation

During the trace serialization, `_PICKLE_BY_VALUE_MODULES` contains the entire stdlib:

```
{'colorsys', 'modulefinder', 'queue', 'pdb', 'collections', 'gzip', 'inspect',
 'decimal', 'enum', 'copy', 'subprocess', 'csv', 'string', 'os', 'typing',
 'json', 're', 'io', 'pickle', 'threading', 'functools', 'pathlib', ...}
```

And cloudpickle's `_should_pickle_by_reference(OrderedDict)` returns `False` because `_is_registered_pickle_by_value(collections)` returns `True`.

## Workaround

```python
from nnsight.intervention.backends.remote import pull_env
pull_env()
import sys
from cloudpickle.cloudpickle import _PICKLE_BY_VALUE_MODULES
_PICKLE_BY_VALUE_MODULES -= sys.stdlib_module_names
```

This forces `pull_env()` to run early, then removes stdlib module names from the set.

## Suggested Fix

In `get_local_env()`, use `sys.stdlib_module_names` (available since Python 3.10) to filter stdlib modules by name rather than by path:

```python
for importer, module_name, is_pkg in pkgutil.iter_modules():
    if module_name in sys.stdlib_module_names:
        continue
    # ... rest of logic
```

This is more robust than path-based filtering.

## Environment

- nnsight 0.6.1
- Python 3.12.9 (Homebrew, macOS 15.4)
- cloudpickle 3.1.2
- torch 2.10.0
- pydantic 2.12.5

## Error Traceback

```
TypeError: cannot pickle 'classmethod_descriptor' object

  File "nnsight/intervention/backends/remote.py", line 529, in request
    data = RequestModel(interventions=interventions, tracer=tracer).serialize(
  File "nnsight/schema/request.py", line 27, in serialize
    data = save(self)
  File "nnsight/intervention/serialization.py", line 964, in dumps
    CustomCloudPickler(buffer, protocol=protocol).dump(obj)
  File "cloudpickle/cloudpickle.py", line 1313, in dump
    return super().dump(obj)
```

keenanpepper:
## Summary

- Fixes remote trace serialization failing with `TypeError: cannot pickle 'classmethod_descriptor' object` on platforms where `sysconfig.get_paths()["stdlib"]` doesn't match the paths returned by `pkgutil.iter_modules()` (confirmed on macOS with Homebrew Python 3.12)
- Replaces the fragile path-based stdlib check with `sys.stdlib_module_names` (available since Python 3.10, nnsight already requires 3.12)
- Removes the now-unused `import sysconfig`

Fixes #619

## Problem

`get_local_env()` used `sysconfig.get_paths()["stdlib"]` to filter stdlib modules by path. On some platforms, this path doesn't match where `pkgutil.iter_modules()` discovers them, so **all stdlib modules** (including `collections`) get misidentified as `"local"` and registered with cloudpickle's `_PICKLE_BY_VALUE_MODULES` via `pull_env()` → `register()`. This causes cloudpickle to serialize `collections.OrderedDict` inline (by value) instead of by reference, which fails because `OrderedDict.fromkeys` is a `classmethod_descriptor` that can't be pickled.

## Fix

Check `module_name in sys.stdlib_module_names` instead of comparing filesystem paths. This is both simpler and platform-independent.

## Test plan

- [x] Verified the minimal reproducer from #619 now works on macOS Homebrew Python 3.12
- [ ] Verify no regression on Linux (where the path-based check may have been working)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tests/test_local_env.py
+++ b/tests/test_local_env.py
@@ -1,0 +1,25 @@
+"""Tests for get_local_env() stdlib filtering."""
+
+import sys
+
+from nnsight.ndif import get_local_env
+
+
+def test_no_stdlib_modules_marked_as_local():
+    """Stdlib modules must never be classified as 'local'.
+
+    Regression test for https://github.com/ndif-team/nnsight/issues/619
+    where a platform-dependent path comparison caused stdlib modules like
+    `collections` to be registered for pickle-by-value, breaking remote
+    trace serialization.
+    """
+    env = get_local_env()
+    packages = env.get("packages", {})
+
+    local_stdlib = {
+        name for name in packages
+        if packages[name] == "local" and name in sys.stdlib_module_names
+    }
+    assert local_stdlib == set(), (
+        f"Stdlib modules incorrectly marked as local: {sorted(local_stdlib)}"
+    )


JadenFiotto-Kaufman:
Looks good to me @keenanpepper thanks.

keenanpepper:
thank Claude Opus 4.6 lmao

Issue #625, opened 2026-03-12
`session(remote=True)` fails for helper-wrapped nested traces with `OSError: source code not available`


asatpathy314:
## Summary
`model.session(remote=True)` fails when a nested `model.trace(...)` is opened inside a helper function.

This works when the nested trace is written inline inside the session body, but fails when wrapped in a helper.

## Minimal reproduction (source: Lasse Jantsch)

```py
import os

from dotenv import load_dotenv
from nnsight import CONFIG, LanguageModel

load_dotenv()
CONFIG.set_default_api_key(os.environ["NDIF_API_KEY"])

model_id = "meta-llama/Llama-3.1-8B"
llm = LanguageModel(model_id, attn_implementation="eager")


def trace(model):
with model.trace("The capital of france is") as tracer:
  pass


with llm.session(remote=True) as session:
trace(llm)

## Actual result

RemoteException: Traceback (most recent call last):
...
⠸ [dd6e3dcf-4a05-48ed-b525-c20c112d8888] RUNNING    (0.0s) Your job has started running.
✗ [dd6e3dcf-4a05-48ed-b525-c20c112d8888] ERROR      (0.3s)
Traceback (most recent call last):
  File "/raid/conda/envs/lmj_nnsight_test/lib/python3.12/site-packages/IPython/core/interactiveshell.py", line 3701, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "/tmp/ipykernel_3103743/266074703.py", line 15, in <module>
    with llm.session(remote=True) as session:

RemoteException: Traceback (most recent call last):
  File "/u/svcndifuser/ndif-deployment/repos/prod/ndif/src/services/ray/src/ray/nn/backend.py", line 30, in __call__
    saves = tracer.execute(self.fn)
  File "/tmp/ipykernel_3103743/266074703.py", line 16, in <module>
    trace(llm.model.layers[0])
  File "/tmp/ipykernel_3103743/266074703.py", line 2, in trace
  File "/u/svcndifuser/miniconda3/envs/service/lib/python3.12/inspect.py", line 1267, in getsourcelines
    lines, lnum = findsource(object)
  File "/u/svcndifuser/miniconda3/envs/service/lib/python3.12/inspect.py", line 1088, in findsource
    raise OSError('source code not available')

OSError: source code not available
```

## Expected result

The helper-wrapped nested trace should behave the same as an inline nested trace inside session(remote=True) and should not fail source capture.

[...]

PR #626, opened 2026-03-12
Fix remote session OSError when helper function contains nested trace

asatpathy314:
## Summary
- Fix `OSError: source code not available` when `model.trace()` is called inside a helper function within `session(remote=True)`
- Register deserialized function source in `linecache` so `inspect.getsourcelines()` succeeds on the remote worker
- Add regression test for the fix

## Problem

When a helper function containing `model.trace(...)` is called inside `session(remote=True)`, the function is serialized and reconstructed on the remote worker. The nested trace then calls `inspect.getsourcelines()` to extract source, but the original `co_filename` doesn't exist remotely, causing:

```
OSError: source code not available
```

This only happens with helper-wrapped traces, inline traces inside the session body work fine.

## Fix

When `make_function()` reconstructs a deserialized function, register its source in Python's `linecache` under the original `co_filename`. This ensures that subsequent `inspect.getsourcelines()` calls (triggered by nested `model.trace()`) can find the source even though the original file doesn't exist on the remote worker. Registration is skipped if the file already exists in the cache (i.e., running locally).

## Test plan
- [x] New test: `test_deserialized_helper_trace_uses_registered_linecache` — serializes a helper function with `model.trace()`, clears linecache, deserializes, calls it, and verifies linecache is populated
- [x] `pytest tests/test_tiny.py --device cpu` — 18/18 passed
- [x] `pytest tests/test_lm.py --device cpu` — 77/77 passed
- [x] `pytest tests/test_serialization_edge_cases.py --device cpu` — 29/29 passed (1 pre-existing failure in unrelated `test_prohibited_module_os`)

### DIFF
--- a/src/nnsight/intervention/serialization.py
+++ b/src/nnsight/intervention/serialization.py
@@ -37,6 +37,7 @@ import dataclasses
 import inspect
 import io
 import linecache
+import os
 import pickle
 import textwrap
 import tokenize
@@ -74,7 +75,7 @@ def _register_source_with_linecache(filename: Optional[str], source: str) -> Non
     if not filename:
         return
 
-    if linecache.getlines(filename):
+    if os.path.isfile(filename):
         return
 
     lines = source.splitlines(keepends=True)
--- a/src/nnsight/intervention/serialization.py
+++ b/src/nnsight/intervention/serialization.py
@@ -36,6 +36,7 @@ Examples:
 import dataclasses
 import inspect
 import io
+import linecache
 import pickle
 import textwrap
 import tokenize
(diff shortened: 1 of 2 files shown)


asatpathy314:
#625

JadenFiotto-Kaufman:
@asatpathy314 I extended what you did here https://github.com/ndif-team/nnsight/pull/633 . problem was, waht if you had two pieces of code from the same file? they would overwrite like in this example:

```py
from nnsight import LanguageModel, CONFIG

# model = LanguageModel("Qwen/Qwen2.5-0.5B")
model = LanguageModel("openai-community/gpt2")


def trace1(model):
    with model.trace("Hello world") as tracer:
        print("trace1")


def trace2(model):
    with model.trace("Hello world") as tracer:
        print("trace2")


with model.session(remote=True) as session:
    session.trace1()
    session.trace2()

```

asatpathy314:
Nice catch! I didn't think about that.

Issue #501, opened 2025-08-20
`tracer.cache()` pickling


AdamBelfki3:
DoD:

- make the `CacheDict` object properly serializable and de-serializable.
- make sure the `-m cache` tests pass with remote execution.

JadenFiotto-Kaufman:
`CacheDict` round-trips cleanly through the new source-based serialization (`src/nnsight/intervention/serialization.py`) — verified with a quick `cloudpickle.dumps/loads` cycle on a populated `Cache.CacheDict`. The serialization rewrite subsumed this DoD. Closing — if cache-mode remote tests still surface an issue, file a fresh bug with the trace.

PR #628, opened 2026-03-13
Fix CacheDict RecursionError during unpickling (#501)

AlliedToasters:
## Summary
- Guard `CacheDict.__getattr__` against access during unpickling, fixing the `RecursionError` reported in #501

## Root Cause

When NDIF returns remote execution results, `_decompress_and_load()` deserializes them via `torch.load()`. Pickle reconstructs `CacheDict` using `__new__` (skipping `__init__`), then restores the dict state. During this process, `__getattr__` is called before `_path` is initialized. Since `__getattr__` accesses `self._path` on its first line, this triggers another `__getattr__` call → infinite recursion.

## Fix

Add a guard at the top of `__getattr__`:

```python
if attr == "_path":
    raise AttributeError(attr)
```

This lets pickle fall back to normal attribute resolution via `__dict__`, which works correctly once `__init__` has run.

## Verification

Isolated reproduction confirms the fix:

```python
import pickle, io

# Simulated CacheDict (dict subclass with __getattr__ accessing self._path)
class CacheDictBroken(dict):
    def __init__(self, data, path=''):
        self._path = path
        super().__init__(data)
    def __getattr__(self, attr):
        path = self._path + '.' + attr  # RecursionError here during unpickle
        raise AttributeError(attr)

cd = CacheDictBroken({'k': 'v'}, path='test')
buf = io.BytesIO()
pickle.dump(cd, buf)
buf.seek(0)
pickle.load(buf)  # RecursionError!

# With guard: succeeds
```

## Test plan
- [x] Isolated pickle round-trip confirms fix resolves RecursionError
- [x] Downstream project (lmprobe) verified remote extraction works with this fix on nnsight 0.6.0 and 0.6.1

### DIFF
--- a/src/nnsight/intervention/tracing/tracer.py
+++ b/src/nnsight/intervention/tracing/tracer.py
@@ -160,6 +160,13 @@ class Cache:
             return dict.__getitem__(self, key)
 
         def __getattr__(self, attr: str):
+            # Guard against access during unpickling: pickle uses __new__
+            # (skipping __init__) then restores state, so _path may not
+            # exist yet when __getattr__ is called.  Without this guard
+            # torch.load() triggers infinite recursion (issue #501).
+            if attr == "_path":
+                raise AttributeError(attr)
+
             path = self._path + "." + attr if self._path != "" else attr
 
             if any(key.startswith(path) for key in self):


JadenFiotto-Kaufman:
Already fixed

Issue #627, opened 2026-03-13
RecursionError when wrapping models that have a child module named output


AntoninPoche:
### Summary

Wrapping some models with `NNsight` / `LanguageModel` crashes at construction time with a `RecursionError`.

This seems to happen when the model contains a child module whose name collides with an `Envoy` mount point such as `output`.

I first hit this through another library (`interpreto`), but the problem reproduces with plain `nnsight`, so this does not appear to be downstream-specific.

### Environment

Reproduced on:
- `nnsight==0.6.2`
- Python `3.12.3`

### Reproduction

Minimal repro with plain `NNsight`:

```python
from nnsight import NNsight
from transformers import BertConfig, BertForMaskedLM

model = BertForMaskedLM(BertConfig())

NNsight(model)
```

Minimal toy example:
```python
import torch
from nnsight import NNsight

class MyClass(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.output = torch.nn.Linear(2, 2)

    def forward(self, x):
        return self.output(x)

NNsight(MyClass())
```

### Expected behavior

Construction should succeed, and nnsight should preserve access to the colliding child module via `.nns_output` as intended.

### Actual behavior
Construction fails with a `RecursionError`. The traceback goes through:

- `Envoy._add_envoy`
- `Envoy._handle_overloaded_mount`
- `Envoy.__setattr__`
- `Envoy.__getattr__`
and eventually loops during attribute access / error formatting. (full error in the end)

### Full error message

```
/home/antonin.poche/interpreto/.venv/lib/python3.12/site-packages/nnsight/intervention/envoy.py:852: UserWarning: Module `model` of type `<class '__main__.MyClass'>` has pre-defined a `output` attribute. nnsight access for `output` will be mounted at `.nns_output` instead of `.output` for this module only.
  warnings.warn(
---------------------------------------------------------------------------
RecursionError                            Traceback (most recent call last)
File ~/interpreto/.venv/lib/python3.12/site-packages/IPython/core/formatters.py:770, in PlainTextFormatter.__call__(self, obj)
    763 stream = StringIO()
    764 printer = pretty.RepresentationPrinter(stream, self.verbose,
    765     self.max_width, self.newline,
    766     max_seq_length=self.max_seq_length,
    767     singleton_pprinters=self.singleton_printers,
    768     type_pprinters=self.type_printers,
    769     deferred_pprinters=self.deferred_printers)
--> 770 printer.pretty(obj)
[...]

PR #629, opened 2026-03-13
Fix RecursionError in Envoy overloaded mount handling for modules with child named output

AntoninPoche:
### Summary

This fixes a constructor-time `RecursionError` when wrapping models that contain child module names colliding with `Envoy` mount points such as `output`.

### Linked issue

The issue is detailed in #627 

### Root cause

`Envoy._handle_overloaded_mount()` dynamically creates a Preserved subclass and currently reassigns the instance class with:
```python
self.__class__ = new_cls
```

However, `Envoy.__setattr__` is overloaded and may forward attribute assignments to the wrapped torch module. In this path, the class reassignment does not behave like a normal Python instance-level `__class__` update and can corrupt the wrapped module state.

That eventually causes recursive attribute access during envoy mounting / error formatting, resulting in a `RecursionError`.

### Fix

Use raw object-level assignment for the class update inside `_handle_overloaded_mount`, so the reassignment is applied to the Envoy instance itself and does not go through `Envoy.__setattr__`.

This preserves the existing overloaded-mount behavior while avoiding the recursive failure.

### Commits

This PR first reproduces the bug in a minimal test in 4860de946afc57b8872aa63f2134b28dd3fca5f6, then fixes it in 11458e460c899024cfb38eb7c3745dd5613b5649 with a one-line change.

### Contributing

I have read the `contributing.md` file and confirm that the tests pass.

This has been mainly driven by GPT-5.4, but I cross-checked every step.

### DIFF
--- a/src/nnsight/intervention/envoy.py
+++ b/src/nnsight/intervention/envoy.py
@@ -866,7 +866,7 @@ class Envoy(Batchable):
                 {},
             )
 
-            self.__class__ = new_cls
+            object.__setattr__(self, "__class__", new_cls)
 
         # Get the normal proxy mount point
         mount = getattr(Envoy, mount_point)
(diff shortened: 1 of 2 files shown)


JadenFiotto-Kaufman:
looks great thanks!

PR #642, opened 2026-03-26
vLLM Intervention Gaps: Compatibility Layer, Deferred Exceptions, and Guide

khaiwang:
# vLLM Intervention Gaps: Compatibility Layer, Deferred Exceptions, and Guide

## Summary

- Transparent HF-compatibility layer so users write the same intervention code on both HF and vLLM backends
- Deferred exception handling so user errors and `tracer.stop()` don't kill the vLLM engine
- Comprehensive VLLM_GUIDE.md documenting every feature's HF pattern, vLLM gap, and status
- Prefix caching disabled by default as a precaution

## HF-Compatibility Layer (`VLLMBatcher`)

vLLM's internal architecture differs from HuggingFace in ways that break standard intervention code. The compat layer (`pre_user_transform` / `post_user_transform` hooks in the interleaver) transparently transforms values so users see HF semantics:

| Gap | Raw vLLM | After compat |
|-----|---------|-------------|
| Decoder layer output | `(mlp_out, residual)` 2-tuple | `(combined,)` 1-tuple |
| `.save()` mutation | Corrupted by `fused_add_rms_norm` | Clone-protected |
| Layer `.input` | int64 position IDs | Float hidden states |
| RMSNorm output | `(normalized, residual)` tuple | Single tensor |
| `RowParallelLinear` output | `(tensor, bias)` tuple | Single tensor |
| `layer.skip()` | Crashes engine | Auto-decomposes for fused norm |
| In-place ops (all models) | `InferenceMode` error | Cloned out of inference mode |

Detection uses init-time heuristics + runtime guards. Verified on Qwen2, Llama, Mistral, Mixtral, Gemma2, DeepSeek-V2, Phi3. GPT-2 correctly excluded from dual-stream transforms but still gets inference-mode cloning.

Also replaces `_transform_stack` (fragile positional list) with `_transform_frames` dict keyed by `(id(module), hook_type)`.

## Deferred Exception Handling

Previously, exceptions in user trace code (including `tracer.stop()`) were re-raised inside PyTorch hooks during the forward pass, unwinding `execute_model()` and permanently killing the vLLM engine.

Fix: `_defer_exceptions` flag on the interleaver. When set by `GPUModelRunner`:
- `handle_exception_event` stores the exception on the mediator instead of re-raising
- The forward pass completes normally
- Exceptions are keyed by request ID in `collect_nnsight()` so each invoke identifies its own error
- `EarlyStopException` is filtered (intentional control flow); real errors are re-raised at the trace boundary

Per-mediator isolation: one failing invoke doesn't kill other invokes' hooks — cancelled mediators are skipped, other mediators keep running.

Tested: stop, single error, one-of-N error, all-error, mixed stop+error — engine survives all.

## Remaining Differences (verified)

These require different vLLM syntax and are not fixable by the compat layer:

| Difference | HF | vLLM |
|---|---|---|
| Logits | `model.lm_head.output` `[1, seq, vocab]` | `model.logits.output` `[1, vocab]` |
| Generation | `model.generate(max_new_tokens=N)` | `model.trace(max_tokens=N)` |
| gate/up proj | Separate modules | `gate_up_proj` merged |
| q/k/v proj | Separate modules | `qkv_proj` merged |
| `__name__` guard | Not required | Required (vLLM uses `spawn`) |
| Module calls outside trace | Works | Fails (weights on meta device) |
| Cross-invoke dependencies | Barriers supported | No barriers (execution order not guaranteed) |

## Hard Blocks

- Gradients: `torch.inference_mode()` disables gradient tracking
- Attention weights: PagedAttention computes in C/CUDA, no Python tensor
- Source tracing on fused modules: trivial Python wrappers only

## Test Plan

- [x] 6/6 exception handling tests pass (test_stop_and_errors.py)
- [x] 19/19 existing vLLM pytest pass
- [x] Compat layer validated on Qwen2-0.5B (save, ablation, steering, patching, skip, logit lens)
- [x] GPT-2 in-place ops work after inference-mode clone fix
- [x] Remaining differences verified with cross-backend comparison tests


src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
```diff
@@ -0,0 +1,612 @@
...
+```python
+with model.trace("Hello"):
+    hidden = model.transformer.h[-1].output[0].save()
+# hidden is the full residual-stream hidden state, shape [batch, seq, hidden_dim]
+```
+
+### vLLM (with compat layer — same code works)
+```python
+with model.trace("Hello", temperature=0.0, top_p=1):
+    hidden = model.model.layers[-1].output[0].save()
```
Butanium: isn't this PR supposed to unwrap the tuple automatically?

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
```diff
@@ -0,0 +1,612 @@
...
+```
+
+---
+
+## Accessing Hidden States
+
+### HF pattern
+```python
+with model.trace("Hello"):
+    hidden = model.transformer.h[-1].output[0].save()
```
Butanium: this is not always true, for some hf version and model you don't need to do .output[0] you can do .output!

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
```diff
@@ -0,0 +1,612 @@
...
+```python
+with model.trace("Hello"):
+    logits = model.lm_head.output.save()       # shape [batch, seq, vocab]
+    last_logits = model.lm_head.output[:, -1]   # last token logits
+```
+
+### vLLM equivalent
+```python
+with model.trace("Hello", temperature=0.0, top_p=1):
+    logits = model.logits.output.save()   # shape [1, vocab_size]
```
Butanium: wait you can't get all the logits of your prompt with vllm? I think it's a certain flag. I guess it can be disabled by default and nnterp could enable it by default

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
```diff
@@ -0,0 +1,612 @@
...
+with model.trace("Hello"):
+    layer_input = model.transformer.h[0].input.save()
+    # Returns float hidden states, shape [batch, seq, hidden_dim]
+```
+
+### vLLM (with compat layer — same code works)
+```python
+with model.trace("Hello", temperature=0.0, top_p=1):
+    layer_input = model.model.layers[4].input.save()
+    # Returns float hidden states, shape [total_tokens, hidden_dim]
```
Butanium: why total_tokens and not seq, are thos differents?

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
```diff
@@ -0,0 +1,612 @@
...
+### HF pattern
+```python
+with model.trace("Hello"):
+    hidden = model.transformer.h[0].output[0].save()
+```
+
+### vLLM (with compat layer — same code works)
+```python
+with model.trace("Hello", temperature=0.0, top_p=1):
+    hidden = model.model.layers[0].output[0].save()
```
Butanium: duplicated with above

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
```diff
@@ -0,0 +1,612 @@
...
+        barrier()
+        model.model.layers[5].output[0][-1, :] = clean_hs          # 2D indexing
+        logits = model.logits.output.save()
+```
+
+The compat layer handles combining/decomposing the dual streams, cloning, and tuple management. Only remaining difference: `[-1, :]` (2D) instead of `[:, -1, :]` (3D), and `model.logits.output` instead of `model.lm_head.output`.
+
+---
+
+## Logit Lens
```
Butanium, line 241: main difference is that you can't do logit lens outisde a trace with vllm while you can with hf (like call the module ioutside of a trace)

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
```diff
@@ -0,0 +1,612 @@
...
+    for step in tracer.iter[:]:
+        logits.append(model.lm_head.output[0][-1].argmax(dim=-1))
+    output = model.generator.output.save()
+```
+
+### vLLM equivalent
+```python
+import nnsight
+
+with model.trace("Hello", max_tokens=3, temperature=0.0, top_p=1) as tracer:
```
Butanium: 5 -> 3?

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
```diff
@@ -0,0 +1,612 @@
...
+    logits = nnsight.save(list())
+    for step in tracer.iter[:]:
```
Butanium, line 288: is this advised instead of tracer.all?
idk what people use these days

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
```diff
@@ -0,0 +1,612 @@
...
+### HF pattern
+```python
+with model.trace("Hello"):
+    normed = model.transformer.ln_f(hidden_states)  # Returns single tensor
+```
+
+### vLLM (with compat layer — same code works)
+```python
+with model.trace("Hello", temperature=0.0, top_p=1):
+    normed = model.model.layers[0].input_layernorm.output  # Returns single tensor
```
Butanium: here you compare vllm layer 0 norm with hf fineal layer norm

Butanium (changes requested):
added some comments on the VLLM_GUIDE.md

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
khaiwang: The default usage case in nnsight's main doc looks like this, not sure whether we want to unwrap it. The pr is solving the problem that vllm separates hidden_states with residual stream, the raw output is (hidden_states, residual), users have to manually add them together to get the output they want.

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
khaiwang: This is mirrored from nnsight's main doc, seems we need to note this concern there?

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
khaiwang: Wow thx for pointing it out. I checked how can we make it, my concern is computing logits for every token might be expensive (just a guess, maybe not that expensive). We will document it here as a future improvement.

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
khaiwang: actually no difference, total_tokens inherits from vllm's continuous batching, worth note it here.

src/nnsight/modeling/vllm/intervention-gaps/VLLM_GUIDE.md
khaiwang: Copied from nnsight's doc. @JadenFiotto-Kaufman do you have any comment for it?

src/nnsight/modeling/vllm/model_runners/GPUModelRunner.py
```diff
@@ -381,6 +386,18 @@ def execute_model(
...
+        interleaver._defer_exceptions = False
+
+        # Safety net: if __enter__ failed or forward was interrupted before
+        # return_value could be assigned, build a minimal valid output.
+        if return_value is None:
+            from vllm.v1.outputs import ModelRunnerOutput
+            req_ids = list(scheduler_output.num_scheduled_tokens.keys())
+            return_value = ModelRunnerOutput(
+                req_ids=req_ids,
+                req_id_to_index={rid: i for i, rid in enumerate(req_ids)},
```
khaiwang, line 401: can we stop the request inference after the mediator ends.

khaiwang:

PR #645, opened 2026-03-29
Fix backward/gradient through batched narrow views

khaiwang:
### Problem

When using `.grad` in a trace with multiple input invokes (batching), backward hooks never fire:

```python
with model.trace() as tracer:
    with tracer.invoke(clean_tokens):
        clean_out = model.transformer.h[-1].output[0].save()
    with tracer.invoke(corrupted_tokens):
        layer_out = model.transformer.h[-1].output[0]
        layer_out.requires_grad_(True)
        logits = model.lm_head.output
        value = vqa_metric(logits)
        with value.backward():
            grad = layer_out.grad.save()  # ValueError: grad was not provided
```

This blocks attribution patching, integrated gradients, and any gradient-based interpretability technique that uses multiple invokes in a single trace.

### Root Cause

`Batcher._narrow()` creates dim-0 narrow views for each invoke's slice of the batch. These views are **not in the autograd forward path** — the full-batch tensor is what flows through the model. PyTorch's `register_hook` on a tensor only fires when gradients are computed for that tensor's autograd node. Since no downstream operation uses the narrow view, its gradient is never computed and the hook never fires.

```
full_output ──────────────→ next_module → ... → loss   (autograd path)
     │
     └── narrow() → view    (dead end — not used by anything leading to loss)
```

A secondary issue: if a user assigns a narrow view back into its base tensor (`module.output = (x,) + rest` where `x` is a narrow view of the output), the in-place assignment in `_swap` creates a self-referential autograd graph that segfaults during backward. This is a PyTorch-level bug triggered by `full[start:end] = view_of_full`.

### Fix

**1. `wrap_grad` (backwards.py):** When registering a backward hook, detect batch-dim narrow views via `tensor._base`. Register the hook on the base tensor (which IS in the autograd path) and narrow the gradient to the invoke's slice inside the hook before providing it to the mediator. (We can discuss this fix, which is the only way that can fix the problem without extra cost. If telling users always reassign the tensor if they need grad is fine, we don't need this fix)

**2. `Batcher._swap` (batching.py):** When `swap_value._base is current_value` (view being assigned back into its own base), route to the `torch.cat` path instead of in-place assignment to avoid the self-referential autograd segfault. swap_value.requires_grad is too weak, if the user code clones or modifies that tensor, autograd graph can capture their gradient correctly. So we only need this self reference guard.

### Known Limitations (documented as TODOs)

- **`_base` view chain fragility:** `tensor._base` always points to the root storage owner, not the immediate parent. If the user further slices a narrow view before accessing `.grad` (e.g. `output[0][:, -1, :].grad`), or if `full_output` is itself a view of an internal buffer, `_base` may skip levels. Uncommon in practice for gradient access patterns.

- **Concurrent mediators (future):** Multiple hooks on the same base tensor would interfere — each `backward()` fires ALL registered hooks, and one finished mediator will remove all of them.

### Tests

10 new tests in `tests/test_backward_batching.py`:
- Gradient in first/second invoke of a batched trace
- Full attribution patching pattern (logit-diff metric)
- Gradient determinism across runs
- Gradient modification (setter) inside backward context
- Single invoke and input+empty invoke regressions
- `retain_graph` with multiple backward passes
- Pop-back pattern (was segfault, now works)
- Normal swap unaffected by the guard

### DIFF
--- a/src/nnsight/intervention/batching.py
+++ b/src/nnsight/intervention/batching.py
@@ -294,6 +294,13 @@ class Batcher:
                 current_value.requires_grad and current_value.is_leaf
             ) or current_value._base is not None
 
+            # When swap_value is a view of current_value (e.g. the user
+            # read a narrow slice from the batcher and assigned it back),
+            # in-place assignment creates a self-referential autograd
+            # graph that segfaults during backward.  Use concat instead.
+            if not needs_concat and swap_value._base is current_value:
+                needs_concat = True
+
             if needs_concat:
                 pre = current_value.narrow(0, 0, batch_start)
                 post = (
--- a/src/nnsight/intervention/tracing/backwards.py
+++ b/src/nnsight/intervention/tracing/backwards.py
@@ -7,6 +7,57 @@ from ..interleaver import Interleaver, Mediator
 from .invoker import Invoker
 
 
+def _get_batch_narrow_info(tensor: torch.Tensor):
+    """If *tensor* is a dim-0 narrow view of a base tensor, return
+    ``(base, start, size)``.  Otherwise return ``None``.
+
+    When multiple invokes are batched, ``Batcher._narrow`` slices the
+    full-batch activation along dim 0 for each invoke.  The resulting
+    view is **not** in the autograd forward path (the full tensor is
+    returned to the next module), so ``register_hook`` on the view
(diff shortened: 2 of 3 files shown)


khaiwang:

PR #647, opened 2026-04-14
Vllm serve combined

khaiwang:
## vLLM: serve mode, intervention gaps, gradient fix

Three independent features rebased into a clean 7-commit history on `dev`.

### nnsight-serve (commits 3–6)

Lightweight single-model server: vLLM does the heavy lifting, nnsight injects intervention hooks on top.

```bash
nnsight-serve Qwen/Qwen3-30B-A3B --port 6677 --gpu-memory-utilization 0.8
```

```python
model = VLLM("Qwen/Qwen3-30B-A3B")  # meta model, no GPU

with model.trace("The Eiffel Tower is in", serve="http://localhost:6677"):
    logits = model.logits.output.save()
```

- **Sync mode** (default): same UX as local -- `.save()` values available after the `with` block
- **Non-blocking mode**: `blocking=False` fires requests in background threads, `t.collect()` returns saves dict. 3.1x speedup with 4 concurrent requests verified.
- **Concurrency safety**: `Globals.saves`/`stack` refactored from class-level attributes to `contextvars` -- each async task gets isolated state. Required because nnsight-serve is the first context where `Globals.enter()/exit()` runs under async concurrency (FastAPI coroutines on the same event loop). Local execution and NDIF are unaffected (backward-compatible via metaclass properties).
- **pyproject.toml**: `nnsight[serve]` extra, `nnsight-serve` entry point, dropped stale `triton==3.5.0` pin

### vLLM intervention gaps (commits 1-2)

- **Deferred exceptions**: intervention errors no longer kill vLLM's engine process. Exceptions are stored per-mediator and surfaced at the trace boundary.
- **13 gaps documented** between vLLM and HF backends with per-gap test scripts and VLLM_GUIDE. Gap 1.1 (in-place mutation corrupts `.save()`) fixed via clone-on-save for inference-mode tensors.
- Semantic gaps (output tuple formats, merged projections, flat batch dims) kept as-is -- nnsight stays architecture-agnostic. Downstream tools like nnterp can bridge model-specific differences.

### Gradient fix (commit 7)

Fix `backward()` through batched narrow views. When multiple invokes are batched, `Batcher._narrow()` creates views not in the autograd forward path -- `register_hook` never fires. Now detects batch-dim views via `_base` and hooks the base tensor instead, narrowing the gradient to the invoke's slice.

### Tests

- `test_serve.py`: 17 tests (inference, interventions, batching, token inputs, non-blocking, numerical match)
- `test_serve_async.py`: 8 tests (timing overlap, result isolation, multi-process, error handling)
- `tests/serve_clients/`: 10 standalone scripts for manual testing with Qwen3-30B-A3B
- Intervention gap tests: 13 vLLM-vs-HF comparison scripts (`run_all.py`)
- `test_backward_batching.py`: gradient through batched invokes
- `test_tiny.py`: 18/18 pass (CPU + GPU), zero regressions

### Commits

```
823858b Fix backward/gradient through batched narrow views
ab0f5e2 Add serve tests, client scripts, and documentation
c420855 Fix serve concurrency and review issues
0550089 Add non-blocking serve mode with concurrent request support
734cd48 Add nnsight-serve: sync single-model server with vLLM backend
a4080d1 Add vLLM intervention gap report and clone-on-save fix
75b8cbf Defer exceptions in vLLM mode to prevent engine death
```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1577,6 +1577,38 @@ with model.scan("Hello"):
 print(dim)  # 768
 ```
 
+### 12. Don't `.save()` Elements Inside a List Comprehension
+
+`.save()` marks an object's `id()` in `Globals.saves`. After the trace, `push` walks the caller's frame locals and only injects variables whose `id()` is in that set. When you `.save()` individual tensors inside a list comprehension, each tensor's id is saved but the tensors aren't bound to any variable name — they're anonymous elements of a list that itself isn't saved. So `push` finds nothing to inject.
+
+```python
+# WRONG — each element is saved but none have names in the frame
+with model.trace("Hello"):
+    results = [model.transformer.h[i].output[0].save() for i in range(12)]
+# NameError: name 'results' is not defined
+
+# WRONG — same problem with a loop appending to a plain list
+with model.trace("Hello"):
+    results = []
+    for i in range(12):
+        results.append(model.transformer.h[i].output[0].save())
+# NameError: name 'results' is not defined
+
+# CORRECT — .save() the list itself, append unsaved elements
+with model.trace("Hello"):
+    results = list().save()
+    for i in range(12):
+        results.append(model.transformer.h[i].output[0])
+# results is a list of 12 tensors
+
+# CORRECT — .save() the comprehension result (the list)
+with model.trace("Hello"):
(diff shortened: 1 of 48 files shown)


khaiwang:
We created a new pr #656 to integrate based on the latest huge refactor (lazy-hook)

PR #652, opened 2026-04-22
Lazy hook execution + eproperty extension API

JadenFiotto-Kaufman:
Reworks nnsight's interleaving architecture around **lazy hooks**: PyTorch hooks are registered on-demand by each mediator and removed after firing, instead of permanent per-module hooks that iterate all mediators on every forward pass.  Also introduces **`eproperty`** — a standardized descriptor for exposing new hookable values in the interleaving system — and adapts Cache, source tracing, the vLLM integration, and generation-step iteration to the new model.

## Headline: performance

On Llama-3.1-8B (single trace, seq_len=102):

| Config | before | after | speedup |
|---|---|---|---|
| `cache` on all modules | 288 ms | 21.8 ms | **13.2x** |
| Single forward pass (no interv) | ~21.5 ms | ~21.5 ms | unchanged |
| `generate` 70 tokens + iter + all layers | ~1505 ms | ~1329 ms | **~12%** |
| `generate` 70 tokens + 2 invokes + all layers | ~1796 ms | ~1561 ms | **~13%** |

The cache case was previously pathological because `Interleaver.handle()` ran `batcher.narrow()` for every module on every forward pass.  Lazy hooks eliminate that completely.

## Architecture changes

- **One-shot hooks**: `hooks.input_hook` / `hooks.output_hook` register PyTorch pre/post hooks only when a mediator actually accesses a module's value.  They `handle.remove()` themselves after firing.  Ordering across mediators is preserved via `add_ordered_hook`, which inserts hooks sorted by `mediator_idx`.
- **Sentinel output hook**: `Interleaver.wrap_module` installs a no-op forward hook so PyTorch stays in the hook dispatch path; otherwise dynamically added hooks during a forward pass would be skipped by PyTorch's zero-hook fast path.
- **Persistent iter hooks**: `IteratorTracer` installs tracker-bumping output hooks with `mediator_idx=float('inf')` for the lifetime of a `for step in tracer.iter[...]:` loop.  These are the sole source of truth for `mediator.iteration_tracker[path]`, which one-shot hooks read to decide when to fire.  Removed in the loop's `finally` block.
- **Persistent cache hooks**: `tracer.cache()` installs output (and optional input) hooks with `mediator_idx=float('inf')` on its target modules for the interleaving session.  Hooks narrow per mediator `batch_group` and feed values straight into `Cache.add(module_path, key, value)` — no provider regex, no iteration parsing.
- **Source tracing**: `OperationEnvoy` now holds its own `pre_hooks` / `post_hooks` / `fn_hooks` / `fn_replacement` lists.  The `wrap` closure inside `Envoy.source` checks `op_envoy._has_hooks()` — untouched operations return `fn` unchanged (zero overhead), touched operations go through `Interleaver.wrap_operation` which processes the op's hook lists.  Recursive source tracing is handled by a one-shot fn-hook + persistent `fn_replacement`.
- **vLLM tensor-parallel hooks**: `VLLMBatcher.wrap` now uses `add_ordered_hook` with explicit `mediator_idx` (pre-hooks `-inf`, post-hooks `+inf`) so its gather/split hooks survive the dict rebuild that happens when mediator hooks are inserted mid-session.

## New extension API: `IEnvoy` + `eproperty`

`eproperty` is now the standardized way to plug a new hookable value into the interleaving system.  Any class that exposes an `interleaver` attribute (optionally a `path`) is an `IEnvoy` and can host `eproperty` descriptors:

```python
from nnsight.intervention.interleaver import IEnvoy, eproperty

class MyExtension(IEnvoy):
    interleaver: Interleaver  # required
    path: str                  # optional; key is used alone if absent

    @eproperty(description="my custom value")
    def my_value(self): ...
```

Reading the property during a trace issues a blocking request via the interleaver; writing schedules a swap.  A model runner on the provider side pushes values via `eproperty.provide(obj, value)`, which fans out to all mediators through `Interleaver.handle`.

All four built-in envoy types already implement `IEnvoy` and use `eproperty`:

- `Envoy`: `.output`, `.input`, `.inputs`
- `OperationEnvoy` (source tracing): `.output`, `.input`, `.inputs`
- `InterleavingTracer`: `.result`
- vLLM `VLLM`: `.logits`, `.samples`

### When you would define your own `eproperty`

The built-in module-level `.output` / `.input` hooks cover the common case of intercepting a PyTorch module's forward pass.  You reach for a custom `eproperty` when the value you want to expose **doesn't** line up with a module boundary and isn't produced by a normal forward hook.

A few scenarios where this pattern earns its keep:

- **Sampler / generation hooks for custom runtimes.**  vLLM is the reference example: `VLLM.logits` and `VLLM.samples` expose values that don't flow through any PyTorch module's forward — they're produced by vLLM's sampler and pushed in via `eproperty.provide`.  If you write an integration for another inference runtime (SGLang, TensorRT-LLM, a custom async server, etc.), you define an `eproperty` for each value you want users to observe or swap, and your runtime calls `provide()` whenever it produces one.
- **Tracer-level values.**  `InterleavingTracer.result` is just a function return value — there is no module to hook.  The tracer pushes the return via `eproperty.provide(tracer, return_value)` inside `Envoy.interleave`.  Any tracer subclass that wants to expose a synthetic "late" value (final loss, perplexity, a post-processing step, an aggregated stat across invokes) can do the same.
- **Streaming / intermediate results.**  Anything produced incrementally — tokens from a streamer, chunks from a decoder, per-step metrics from a scheduler — can be wrapped in an `eproperty` so user intervention code can access it with the same `.save()` / swap semantics as module outputs.
- **Diffusion / multi-stage pipelines.**  Values that live between stages (e.g. latents between denoising steps, intermediate conditioning signals) are natural candidates — the pipeline code calls `provide()` between stages.
- **Auxiliary modules you want to expose as first-class values.**  If your model has a submodule whose "output" is really a synthesized value (SAE features, router logits, KV-cache reads), you can back an `eproperty` with a `@requires_output` decorator and register a one-shot hook the same way `Envoy.output` does — but present it under a custom name and with your own pre/post-processing.

Post-processing and pre-processing on read/write are available via decorators, mirroring `@property.setter`:

```python
@eproperty(key="input")
def input(self): ...

@input.preprocess
def input(self, value):
    # called on __get__ after the interleaver returns — extract a view
    return [*value[0], *value[1].values()][0]

@input.postprocess
def input(self, value):
    # called on __set__ before the interleaver swap — repack into full form
    inputs = self.inputs
    return (value, *inputs[0][1:]), inputs[1]
```

Multiple `eproperty` instances can share the same `key` to offer different views (read-only convenience, pre-processed, etc.) of the same underlying provider.  The `description` argument controls whether the property appears in the module repr tree.

## User-facing API changes

- `Envoy._interleaver` → `Envoy.interleaver` (public, required by `IEnvoy`).
- `OperationEnvoy.name` removed; `.path` is now the canonical identifier.
- `Cache.add` signature: `add(provider_string, value)` → `add(module_path, key, value)`.
- `Cache.CACHE_PROVIDER_RE` removed (no more provider-string parsing).
- `Mediator.OutOfOrderError` renamed to `Mediator.MissedProviderError`.
- `Interleaver.handle` now returns a value (the final mediator's handled value) — used by `operation_fn_hook` for recursive source tracing.

---

## Additional changes since this PR was opened

A few follow-ups landed on top of the original lazy-hook + eproperty work in this branch. Most are internal / surface-level, none change the headline architecture above.

- **`.source` refactor.** Split per-module `SourceAccessor` (cached on `module.__source_accessor__` so it survives `module.forward` replacement) from per-Envoy `SourceEnvoy` / `OperationEnvoy` wrappers. Multiple Envoys / Interleavers now share one accessor per module instead of each rebuilding the AST injection.
- **`.source` works under `tracer.iter[...]`.** Iteration tracker bumping now recurses through op-level paths so per-step source-tracing is correct in multi-token generation.
- **`@hooked_output` / `@hooked_input` sugar decorators.** Combine `@eproperty()` + the matching `@requires_*` into one decorator (with operation-level variants `@hooked_operation_*`). The lower-level `@eproperty()` + `@requires_*` form is still valid for advanced cases (custom `requires_*`, runtime-pushed eproperties like `VLLM.logits`). Built-in usages (`Envoy.output`/`.input`/`.inputs`, `OperationEnvoy` equivalents) switched to the sugar form.
- **`IEnvoy.path` is now `Optional[str]`.** The Protocol declaration matches the runtime semantics — tracer-level eproperties without a path prefix were always supported, the type was just wrong.
- **`async_backend.py` clarification.** Async-mode saves are collected only when `output.finished == True`. The vLLM README description is updated to match. Per-yield streaming is noted as a planned future option.
- **Documentation restructure (out of band).** `CLAUDE.md` is now a thin (~180 line) router that funnels agents to a new `docs/` tree (~100 recipe-style files: concepts, usage, models, remote, patterns, gotchas, errors, developing, reference). Doc-vs-source drift on `transformers >= 5.0` (block outputs are tensors, not tuples) was swept across the docs in one pass.
- **Agent-eval suite expanded.** `tests/agent-evals/` now covers 65 tasks (33 code-generation + 32 multiple-choice), with selectable doc bundles (`minimal` / `router` / `full` / `legacy`), a static / browse mode toggle (browse mode gives the agent a `Read` tool scoped to the chosen bundle), and a `claude-code` provider that shells out to the local `claude` CLI. Used to benchmark this PR's doc + API changes — see `tests/agent-evals/report.md` for findings.


Butanium:
@JadenFiotto-Kaufman this looks cool! any idea about the test that fail on this PR.
Also i'm not very familiar with nnsight internals so if there is any part of the code you'd want feedback on / speicifc synthax / design choice?
@claude could you provide some code snippet to show how the xchanges will change nnsight code?

gsarti:
My understanding from reading the updated .md files is that most user workflows should remain the same, given that the lazy hooking/unhooking happens behind the scenes. Am I right @JadenFiotto-Kaufman?

src/nnsight/intervention/interleaver.py
```diff
@@ -123,96 +422,92 @@ def initialize(
...
+           present, the module's original forward is bypassed and the skip
+           value is returned directly.
+        2. A **sentinel output hook** — an empty ``register_forward_hook``
+           that returns ``output`` unchanged.  This is required because
+           PyTorch's ``Module.__call__`` fast-paths when *no* hooks are
+           registered: if a module has zero hooks at call time, dynamically
+           added hooks during the forward pass will never fire.  The sentinel
+           ensures PyTorch always goes through the hook dispatch path so that
+           one-shot hooks registered by :func:`hooks.input_hook` and
+           :func:`hooks.output_hook` can be picked up mid-forward.
```
khaiwang, line 509: Is there any example that a hook registration can happen after pytorch has finished the hook check so we need this sentinel output hook?

src/nnsight/intervention/envoy.py
```diff
@@ -1178,24 +1246,45 @@ def __init__(
...
+
+    def _has_hooks(self) -> bool:
+        """Return True if any hooks or a fn replacement are installed.
+
+        Used by the ``wrap`` closure in :attr:`Envoy.source` (and the
+        nested one in :attr:`OperationEnvoy.source`) to decide whether
+        an operation needs wrapping or can be called directly.
+        """
+        return bool(
+            self.pre_hooks or self.post_hooks or self.fn_hooks or self.fn_replacement
```
khaiwang: I can't find where do we reset fn_replacement after one request is finished. The hook will be persistent if this reset logic is missing once a request has set this fn_replacement.

JadenFiotto-Kaufman:
> @JadenFiotto-Kaufman this looks cool! any idea about the test that fail on this PR.
> Also i'm not very familiar with nnsight internals so if there is any part of the code you'd want feedback on / speicifc synthax / design choice?
> @claude could you provide some code snippet to show how the xchanges will change nnsight code?

Yeah when I ping you a PR I don't necessarily mean you need to really deep dive and understand. Just making sure you're aware of the more intensive changes. More than the last hooking, the eproperty might be worth incorporating more into nnterp if there are certain transformations you want on module inputs / outputs. Ill have an nnterp PR using eproperty for making hooking the attention by head cleaner @AdamBelfki3. I'll share a snippet here when I get home

JadenFiotto-Kaufman:
> My understanding from reading the updated .md files is that most user workflows should remain the same, given that the lazy hooking/unhooking happens behind the scenes. Am I right @JadenFiotto-Kaufman?

Yeah no breaking changes for the core nnsight API.

src/nnsight/intervention/tracing/iterator.py
```diff
@@ -1,34 +1,190 @@
...
     def __getitem__(self, iteration: Union[int, slice, list[int]]):
-        return IteratorTracer(iteration, self.interleaver)
+        return IteratorTracer(iteration, self.interleaver, self.model)
+
+
+def _register_iter_hooks(mediator, model) -> List:
+    """Register persistent hooks that bump ``mediator.iteration_tracker``
+    for every module after each forward pass.
+
+    These hooks are the single source of truth for the per-mediator
```
khaiwang, line 99: Why we need this counter to be per-module? A forward pass for a mediator is always lock-stepped, I cannot think up of an example where different modules for a mediator are in different steps. If I'm right, we can set one counter at the outmost module and have sth like a mediator.step_counter, avoid iterating all modules for each generation request.

Butanium:
Thanks!

Butanium (commented):
Excited to see this added to nnterp (forked from the dev branch would be nice)

src/nnsight/intervention/interleaver.py
JadenFiotto-Kaufman: Yeah take this example:

```py
with model.trace(...):

    x = model.layers[0].output
    y = model.output
```

Nnsight will get to the layers [0] line wait for layer 0 output and add a hook on layer 0. Model will begin executing. Layerr 0 will be hit And nnsight will resume, reach model.output. we add a hook on the root module, but we're already in the forward pass of the root module to get to the layer 0. Make sense?

src/nnsight/intervention/tracing/iterator.py
JadenFiotto-Kaufman: I mean that's just not true in the general case right. Take a state space model like Mamba which loops over a module in every layer. So it might be the first time the  layer is called but the 10th time the inner module is called.

khaiwang:
Makes sense. So are users expected to compute which iteration they want of that specific module and compute the relationship of inner and outer loop? (Sry I don’t quite know how people interact with modules like mamba)

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

src/nnsight/intervention/tracing/iterator.py
khaiwang: (Just out of curiosity, should not resolve in this PR) Do you think this can be optimized to a global counter for general cases and apply this counter to specific modules that have this intra-forward pass loops (maybe configured in a config file so we can adopt quickly for new models), and does that worth it? Since we expect more workloads having multi-token generation, what the cost of having this iteration counter for them.

PR #656, opened 2026-04-29
Vllm serve

khaiwang:
## nnsight-serve

Lightweight single-model server: vLLM does the heavy lifting, nnsight injects intervention hooks on top.

```bash
nnsight-serve Qwen/Qwen3-30B-A3B --port 6677 --gpu-memory-utilization 0.8
```

```python
model = VLLM("Qwen/Qwen3-30B-A3B")  # meta model, no GPU

with model.trace("The Eiffel Tower is in", serve="http://localhost:6677"):
    logits = model.logits.save()
```

- **Sync mode (default)**: same UX as local — `.save()` values available after the `with` block.
- **Non-blocking mode**: `blocking=False` fires requests in background threads; `t.collect()` returns the saves dict. 3.1× speedup with 4 concurrent requests verified.
- **Concurrency safety**: `Globals.saves` / `Globals.stack` route through `contextvars` so each FastAPI handler coroutine gets isolated state. Required because nnsight-serve is the first context where `Globals.enter()/exit()` runs under async concurrency on a shared event loop. Local execution and NDIF are unaffected — the metaclass preserves the public attribute API.
- **Per-request save namespacing**: worker `collect_saves` returns `{base_id: {var_name: value}}` instead of a flat dict. Without this, two concurrent independent traces whose user code uses overlapping variable names (`logits`, `x`, …) would clobber each other at the engine output level.
- **`ServeInterleavingTracer` subclass**: holds the in-flight HTTP future and exposes `.collect()` as a typed attribute on a dedicated subclass — keeps serve-specific state out of the base tracer (selected via `tracer_cls=` from `VLLM.trace`).
- **`pyproject.toml`**: `nnsight[serve]` extra, `nnsight-serve` console entry point.

## vLLM intervention gaps

- **Deferred exceptions**: intervention errors no longer kill vLLM's engine process. The interleaver enters defer mode around `execute_model` / `_sample`, exceptions are stored per-mediator on `_deferred_exception`, and the engine ships them back to the client as `__nnsight_exceptions__: {base_id: {type, message}}` alongside any saves that were already collected. The engine stays alive for subsequent requests instead of `EngineDeadError`-ing on the next one.
- **13 gaps documented** between vLLM and HF backends (`intervention-gaps/REPORT.md`, `VLLM_GUIDE.md`) with per-gap test scripts (`run_all.py`). Semantic gaps (output tuple formats, merged projections, flat batch dims) kept as-is — nnsight stays architecture-agnostic. Downstream tools like nnterp can bridge model-specific differences. In-place mutation on inference-mode tensors is documented as a user-handled case (`output[:] = 0` is unsupported on vLLM-produced tensors; users clone first).

## Gradient fix for transformers

Fix `backward()` through batched narrow views. When multiple invokes are batched, `Batcher._narrow()` creates views that are not in the autograd forward path — the full-batch tensor is what flows through the model, so `register_hook` on the view never fires. Now detects batch-dim narrow views via `_base` and hooks the base tensor instead, narrowing the gradient to the invoke's slice inside the hook (and splicing user-supplied replacements back into the full grad). Also guards `Batcher._swap` against self-referential view assignment, which corrupts the autograd graph.

## Tests

| Suite | Result |
|---|---|
| `tests/test_serve.py` | 15/17 pass against live nnsight-serve (the 2 failing are intentional in-place inference-tensor cases) |
| `tests/test_serve_async.py` | 8/8 — timing overlap, result isolation across concurrent traces with overlapping save names, multi-process clients, error handling |
| `tests/serve_clients/` | 10 standalone scripts for manual testing with Qwen3-30B-A3B |
| `intervention-gaps/run_all.py` | 13 vLLM-vs-HF comparison scripts |
| `tests/test_backward_batching.py` | gradient through batched invokes |
| `tests/test_tiny.py` | 18/18 pass on CPU |

## Notes on the dev rebase

The branch was originally on the previous architecture; this version sits on top of lazy-hook + eproperty refactor (PR #652). The rebase was clean for this PR's behavior — every fix above is preserved, the per-trace saves and contextvars wiring is intact, and the engine-survival path was retargeted at dev's interleaver internals without functional changes.

One user-facing rename came along with the rebase: `VLLM.logits` / `VLLM.samples` / `VLLM.generator` are now `eproperty`-backed tensors, accessed as `model.logits.save()` instead of `model.logits.output.save()`. The serve tests, client scripts, and examples in this PR have been updated accordingly.


src/nnsight/modeling/vllm/model_runners/GPUModelRunner.py
khaiwang: then these reference need to be renamed to defer_exceptions. (Maybe we can make defer_exceptions = True by default?)

PR #658, opened 2026-05-03
handle exceptions in converted functions

lmjantsch:
## The Problem

Exceptions in code compiled by `convert()` (which injects operation wrapping into forward methods using generic \<nnsight> filename) crashed `ExceptionWrapper._collect_frames()` with a `KeyError`. The exception handler only knew how to map `Tracer.Info` frames, not `convert()` frames.

## The Solution

Three files were modified:

### 1. `globals.py`
Added a metadata registry:
```python
class Globals:
    converted_fn_files = {}  # Maps <nnsight {metadata_key}> filenames to (co_filename, co_firstlineno, co_name)
```

### 2. `inject.py` 
Modified `convert()` to register metadata before compiling:
```python
metadata_key = hash((fn.__code__.co_filename, fn.__code__.co_firstlineno, fn.__code__.co_name))
filename = f"<nnsight {metadata_key}>"
Globals.converted_fn_files[filename] = (
    fn.__code__.co_filename,
    fn.__code__.co_firstlineno,
    fn.__code__.co_name
)
```

### 3. `util.py`
Updated `ExceptionWrapper._collect_frames()` to check both mappings:
```python
if fname in filename_mapping:
    # Handle Tracer.Info frames (existing logic)
    ...
if fname in Globals.converted_fn_files:
    # Handle convert() frames (new logic)
    co_filename, co_firstlineno, co_name = Globals.converted_fn_files[fname]
    line_number = co_firstlineno - 1 + lineno
    # Reconstruct original location
    ...
else:
    # Unknown frame - show as-is
    frames.append((fname, lineno, name, "", False))
```

## Result

- Exceptions in `convert()` compiled code no longer crash
- Tracebacks now show real file paths and correct line numbers
- Both `Tracer` and `convert()` frames are handled

### DIFF
--- a/src/nnsight/intervention/source.py
+++ b/src/nnsight/intervention/source.py
@@ -80,6 +80,7 @@ from .hooks import (
     hooked_operation_output,
 )
 from .interleaver import IEnvoy, Interleaver
+from .tracing.globals import Globals
 
 if sys.version_info >= (3, 9):
     _ast_to_source = ast.unparse
@@ -191,7 +192,17 @@ def convert(fn: Callable, wrap: Callable, name: str):
     global_namespace.update(module_globals)
     global_namespace["wrap"] = wrap
 
-    filename = "<nnsight>"
+    metadata_key = hash((
+        fn.__code__.co_filename,
+        fn.__code__.co_firstlineno,
+        fn.__code__.co_name,
+    ))
+    filename = f"<nnsight {metadata_key}>"
+    Globals.converted_fn_files[filename] = (
+        fn.__code__.co_filename,
+        fn.__code__.co_firstlineno,
+        fn.__code__.co_name
+    )
 
     if isinstance(tree, ast.Module):
         code_obj = compile(tree, filename, "exec")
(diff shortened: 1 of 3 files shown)


lmjantsch:
I am not sure if saving the file metadata to `tracing.Globals` is the right way to go. If you have a better idea on where to attach the metadata to let me know and I will give it a shot.

lmjantsch:
@AdamBelfki3 @JadenFiotto-Kaufman It would be nice, if this could make its way in the next version :)

JadenFiotto-Kaufman:
Hey — sorry it's been a while since this was posted, and thank you for it! Merging it in. 🙏

Heads up that how we handle exceptions and source rewriting has changed quite a bit in **0.8**: the `ExceptionWrapper` traceback machinery this patched has been replaced (by a simpler `clean_traceback`), and instrumented forwards are now compiled against the module's real filename rather than a synthetic `<nnsight>` one — so the specific `KeyError` here can't occur there and this doesn't port over directly. It did surface the same underlying concern in 0.8 (an exception inside an instrumented `forward` reported a drifted line number), and it inspired an equivalent fix on the 0.8 branch:

https://github.com/ndif-team/nnsight/commit/2fd034f77dd207c242e3be1c1594ce1efa8589a7

Thanks again for flagging it! 🚀

PR #667, opened 2026-05-20
Add Function Vectors reproduction example notebook

Devansh-Ruia:
Adds examples/function_vectors_reproduction.ipynb, which reproduces the function vector extraction from Todd et al. (ICLR 2024) using nnsight's tracing API.

## What this does
- Loads [model name you used], extracts a function vector for the [task name] task using N in-context demonstrations
- Injects the FV into a zero-shot context and measures the task-completion delta
- Documents one place where the nnsight port deviates from the original PyTorch hook implementation

## Why
The repo has no published example linking a paper's specific result to nnsight idioms. This notebook closes that gap for Function Vectors and gives new users a reference for the trace-and-intervene pattern on real interpretability work.

## How it was tested
- Runs end-to-end on a clean kernel restart
- All cells produce output
- Tested on Collab A100

Happy to revise based on review feedback. First contribution to the repo.


JadenFiotto-Kaufman:
Hi @Devansh-Ruia, and sorry this sat since May without a reply. That is our
fault, and a bad way to treat a first contribution.

The awkward part is that a Function Vectors tutorial already exists, just not in
this repo. It is on the docs site under Mini Papers:
https://nnsight.net/tutorials/mini-papers/todd_function_vectors/. It was adapted
from Callum McDougall's ARENA material, with Eric Todd's paper as the source. The
notebook itself lives in https://github.com/ndif-team/nnsight-website at
`docs/tutorials/mini-papers/todd_function_vectors.ipynb`.

So the gap you set out to close is already closed, and notebooks live in the
website repo rather than this one. There is no way you could have known that from
here, and your reason for wanting it was right.

If you read the existing one and think something is missing, a PR against
nnsight-website would be welcome. The thing I would most want written down is
where the nnsight port diverges from the original hook implementation, which you
mention documenting. Only someone who has actually done the port notices that.

A few things to know before you do, since some of them would have blocked this PR
anyway.

Most of the diff is not the notebook. `uv.lock` is 6,870 lines and would pin
everyone's dependency resolution to your machine. `py_mount.cp312-win_amd64.pyd`
and the cp313 one are compiled binaries that should not be committed. There is
also a stray `function_vectors` entry, and 59 lines of `CLAUDE.md` changes I
could not place.

The description still has its template placeholders, "[model name you used]" and
"[task name]", so it does not say what it reproduces.

It targets `main`, but development happens on `0.8`. Our own `CLAUDE.md` tells
contributors to use `dev`, which is wrong and has misdirected several people. We
are fixing that.

The API changed in the 0.8 rewrite, so a notebook written against 0.7 will not
run unmodified. `LanguageModel` is now `TransformersModel`, `.value` is gone, and
generation results come off `tracer.result`.

Closing this as already covered. Do take a look at the existing tutorial.

PR #670, opened 2026-06-05
Pipeline-parallel intervention support for the vLLM backend

khaiwang:
# Pipeline-parallel intervention support for the vLLM backend

**Branch:** `pp-on-dev` → `dev` (42 commits)

## Summary

Adds transparent nnsight intervention support under vLLM **pipeline parallelism**
(PP ≥ 2). A single-GPU-style trace runs unchanged across pipeline stages — read,
modify, and save activations that live on *any* stage, in both single-forward and
multi-token generation, on single-node multiproc **and** multi-node Ray. The user
never writes `if pp_rank == X`; the system makes a cross-stage data dependency
(`layers[5]` on stage 0 feeding `layers[60]` on stage 1) just work.

This PR brings the core design, a series of correctness fixes across intervention
patterns and model architectures, a full performance characterization with one
latency-stall fix, multi-node (Docker-Ray/TCP) validation, and the design docs.

## The problem

PP splits the model by layer across stages; vLLM builds all layer slots on every
rank but only its own slice is real (`PPMissingLayer` stubs elsewhere). nnsight ships
and runs the **same** intervention body on every rank, and a normal module access
**blocks** until that module's forward hook fires. On a stage that doesn't own the
module, the hook never fires → the mediator deadlocks on the first cross-stage line.
The naive "push all intervention state to the owning rank" fix is impossible: the
mediator is a long-lived, stateful worker thread (frame locals, accumulating saves,
loop counters) that can't be serialized mid-flight, and most interventions have no
cross-stage dependency at all.

## The design

Detect "this module lives on another stage" **at the access site, before blocking**,
and return a `LazyRemoteTensor` placeholder instantly. Writes/saves to it are no-ops
(the real write lands on the owning rank running the same line); only a genuine
*consume* materializes it — a pull over a dedicated gloo group from the owning rank's
buffer. Because remote accesses never block, each rank's worker **runs ahead** of its
forward and parks at the first module it owns; a **readiness gate** synchronizes
worker and forward, and a **finalize drain barrier** keeps each rank's serving buffer
alive until every rank's worker has finished pulling.

Full as-built spec: `docs/developing/pp-design.md`. Picture-first intro:
`docs/developing/pp-pipeline-parallelism.md`.

## What's included

**Core PP path**
- `PPEnvoy` / `pp_eproperty` short-circuit on PP-non-local `.output`/`.input` (runs
  regardless of `interleaving`, so downstream/post-teardown accesses don't raise).
- `LazyRemoteTensor` — metadata proxy; no-op writes/saves; materialize-on-consume;
  child-lazy `__getitem__` (correct tuple-element reads) and `__iter__`/`__len__`.
- `PPListener` — per-rank gloo listener; self-identifying single-message requests,
  per-pull response tags (concurrent pulls), check-and-park serving + `dispatch_parked`.
- `PPModuleMap` — module → owning-stage resolution (layer ranges; first/last-rank
  modules incl. `logits_processor` for arches that build it on every rank).
- Run-ahead worker + `_pp_wait_for_mediators` readiness gate; `pp_hook_buffer` clone
  sidecar; save collection + position-wise merge across ranks.

**Correctness fixes (across Qwen2.5/Qwen3/Llama/Gemma2, PP=2)**
- Cross-stage `logits`/`samples` pull resolved the owner from the full module key,
  not the root path (fixes a logits-consume-in-`iter` hang).
- Deep-clone of tuple/list hook values into the buffer (fixes silently-wrong reads of
  upstream decoder-layer tuple outputs; bare-tensor clone left tuple tensors aliased
  to reused forward storage).
- Request-scoped iteration tracker + shape-on-wire ("legacy") pull sizing under
  run-ahead (consumer can't predict the produced token count).
- Concurrent cross-rank pulls via per-pull response tags; non-blocking listener
  (park instead of block) to avoid head-of-line cross-stage deadlock.
- Batched multi-invoke saves: collect saved vars used only by non-first invokes;
  `__iter__`/`__len__` so `tuple(lazy)` doesn't spin on the non-owning rank.

**Performance**
- Characterized end-to-end (Qwen2.5-7B, PP=2) vs vanilla vLLM and vs PP=1, on
  single-node multiproc and 2-node Docker-Ray/TCP:
  - Plain generation is a PP **win** under `enforce_eager` (PP=2 ~25–33% faster than
    PP=1); nnsight tracks vanilla vLLM with no intervention-free penalty.
  - Neither the hook-buffer clone (~2–3 ms/layer, overlaps compute) nor the
    cross-stage pull (~3.4 ms local / ~7 ms TCP) is a fundamental bottleneck.
- **Fix:** reading many cross-stage layers in one forward (logit-lens / save-every-
  layer) hit a fixed **~5 s stall** on the multiproc/async-scheduling path — the
  producer rank cleared its serving buffer before the consumer finished pulling, so
  late pulls blocked until a 5 s `join` timeout (correctness was unaffected; the merge
  backfills). Closed with a request-finalize **drain barrier**
  (`PPListener.drain_barrier`): `save_all_consume` **5.35 s → 0.37 s**, no plain-gen
  regression, deadlock-free (runs under `collective_rpc` with identical
  `finished_req_ids` on exactly the pull group's TP-rank-0 members).

**Multi-node**
- Validated on a 2-node Docker-Ray cluster with `NCCL_P2P_DISABLE`/`NCCL_SHM_DISABLE`
  (cross-stage forced over TCP): correctness and no drain-barrier deadlock; plain gen
  ~2× slower than single-node multiproc (per-token inter-stage transfer over TCP +
  async scheduling off), no cross-stage-read stall.
- ≥3-node Ray PP fails at kv-cache init — a **vLLM** bug (`global_rank` not
  synced in `adjust_rank`; upstream #41287 / PR #41298), reproduced on vLLM without nnsight. Documented and skipped (`docs/developing/pp-multinode-ray-init-bug.md`).

**Docs**
- `docs/developing/pp-design.md` — detailed as-built design spec
- `docs/developing/pp-pipeline-parallelism.md` — illustrated walkthrough + figures.
- `docs/developing/pp-stress-findings.md` — intervention-pattern correctness/perf
  findings (P1/P2, the in-place-write design choice, the stall fix).

## Validation

- Differential PP=1-vs-PP=2 classifier (`/tmp/pp_stress`): 7/7 intervention patterns
  PASS with the drain barrier — `clean`, `read_all` (reldiff 0.0000), `read_split`,
  `cross_consume`, `cross_write`, `ablation`, `steering`.
- Cross-architecture dense validation at PP=2 (Qwen2.5-0.5B/7B/14B, Qwen3, Llama/
  DeepSeek-8B, Gemma2-9B).
- Multi-node: Docker-Ray PP=2 (incl. PP=2×TP=2) correctness + no-deadlock.
- Perf: full benchmark matrix (vanilla/nnsight × pp1/pp2 × multiproc/2-node).

## Notes / known limitations

- `enforce_eager=True` and `enable_prefix_caching=False` are forced (prefix caching
  would skip hooked forwards and silently drop interventions).
- In-place writes on vLLM inference tensors (`module.output[0][:] = …`) are
  intentionally unsupported; use replacement (`module.output = value`).
- `pp_hook_buffer` accumulates per-token clones within a request (scoped-cleared at
  finish); very long generations with many hooked modules may eventually want eviction.

## Follow-ups

- **Real multi-machine (cross-host) evaluation.** The 2-container Ray/TCP test is a
  pessimistic same-host proxy; real hardware differs in the network fabric (vLLM's
  inter-stage transfer is NCCL and can use IB/RoCE RDMA, but nnsight's cross-stage
  pulls use **gloo/TCP** and won't get RDMA acceleration), plus deployment plumbing
  (`*_SOCKET_IFNAME` on the data NIC, routable head IP + open Ray/gloo ports, shared
  FS / identical nnsight+vLLM+weights on both nodes). The pull-heavy cross-stage-read
  workloads are where physical distance shows up most. To be measured separately.
- Optional `pp_hook_buffer` eviction for very long generations.


khaiwang:
close this pr because we have new one based on 0.8

PR #671, opened 2026-06-07
Fix MissedProviderError in .backward() with multiple invokers (#664)

### LINKED ISSUE #664: Bug: MissedProviderError in `.backward()` with multiple invokers
# Bug: MissedProviderError in `.backward()` with multiple invokers

## Reproducing example

```python
with model.trace() as tracer:
    with tracer.invoke(inp_str):          # invoker 1
        out = model.lm_head.output.save()

    with tracer.invoke(inp_str):          # invoker 2
        attn_outs = []
        for layer in model.transformer.h:
            x = layer.attn.c_proj.output
            attn_outs.append(x)

        logits = model.lm_head.output
        loss = logits.sum()

        with loss.backward():
            for x in reversed(attn_outs):
                print(x.grad)            # MissedProviderError here
```

Works fine when only invoker 2 is present (the `.backward()` block is removed or invoker 1 is dropped).

## What goes wrong

With a **single invoker**, `Batcher.narrow()` returns the original module output tensor unchanged (`needs_batching = False`). That original tensor is part of the computation graph that produced `loss`, so PyTorch runs through it during `loss.backward()` and the hook fires.

With **two or more invokers** sharing an input, `needs_batching = True` and `Batcher.narrow()` returns `acts.narrow(0, batch_start, batch_size)` — a **storage-sharing view** of the original tensor. This view is what gets stored in `attn_outs`.

The backward hook is registered on this view. However, the view is **never used as an input to any operation that produced `loss`**. The computation graph for `loss` flows through the original `c_proj_output` tensor (the full-batch output produced during the forward pass), not through the narrowed view that was created purely to deliver a batch slice to the worker thread.

Because the view has no path in `loss`'s backward graph, PyTorch never runs a gradient node for it. The hook never fires.

## Proposed fix

Before registering the backward hook, check whether the tensor is a **storage-sharing view** (i.e., `tensor._base is not None`). If it is, derive the view's position in the parent tensor from standard tensor metadata — storage offset, stride, and size along the batch dimension — and register the hook on the **parent tensor** instead. When the hook fires, narrow the incoming gradient to the correct slice before delivering it to the `BackwardsMediator`. If the user modified the gradient (a swap event), splice the modified slice back into the full parent gradient before returning it to PyTorch.

The provider/requester key (`id(tensor)`) stays tied to the **original view object**, so the `BackwardsMediator`'s existing wait logic requires no changes.

## Open question / potential issue

The fix assumes that whenever the tensor delivered to the worker is a view, the hook should be on the **parent** (base tensor). This is correct for the `Batcher.narrow()` case, because the view was created only for slice delivery and was never fed into a computation leading to `loss`.

However, consider a case where the user themselves constructs `loss` from a narrowed view:

```python
with tracer.invoke(inp_str):
    m_out_last_token = model.transformer.h[-1].output.narrow(1, -1, 1) # select last token hidden_state
    loss1 = loss_fn1(m_out_last_token)

    logits = model.lm_head.output
    loss2 = loss_fn2(logits)

    loss = loss1 + loss2

    with loss.backward():
        print(logits.grad) # works fine
        print(m_out_last_token.grad) # would get the gradient from model.transformer.h[-1].output which is the combined partial gradient through both losses.
```

azrabano23:
## Summary

Fixes #664 — `MissedProviderError` when accessing `.grad` inside a `.backward()` block in an invoke when **two or more invokes share an input**.

## Root cause

With a single invoke, `Batcher.narrow()` returns the module-output tensor unchanged, and that tensor is in the loss graph, so the backward hook fires. With **two or more invokes**, `needs_batching` becomes true and `Batcher._narrow()` returns a storage-sharing view (`acts.narrow(0, start, size)`) for slice delivery. That view is **never an input to any op producing the loss** — the loss flows through the full-batch parent — so PyTorch never runs a grad node for the view and the hook registered on it never fires.

## Fix

- `Batcher._narrow` tags each batch-slice view it creates with `_nnsight_batch_group`.
- `wrap_grad.wrap` detects a tagged view and registers the hook on the **parent** (`tensor._base`), which *is* in the loss graph. When the hook fires it recovers this invoke's slice from the parent gradient with `grad.as_strided(view.shape, view.stride(), view.storage_offset())`, hands it to the mediator, and — if the user edited the gradient (a swap) — splices the edit back into the parent gradient.

Using the view's own `as_strided` geometry (rather than `narrow(0, …)`) is necessary because the parent is the **flattened** `(batch*seq, hidden)` activation, not `(batch, seq, hidden)` — so a naive dim-0 narrow selects the wrong rows.

The redirect is gated on the Batcher's tag, so it applies **only** to batcher-created views. User-constructed views keep the previous behaviour (hook on the tensor directly), which sidesteps the ambiguity raised in the issue's "open question" about user-built views that participate in the loss directly.

## Testing

Added two regression tests to `TestGradients` in `tests/test_lm.py`, run on CPU:

- `test_backward_with_multiple_invokers` — reproduces the issue scenario and asserts the batched-invoke gradient (a) is provided and (b) **numerically equals** the single-invoke gradient for the same input.
- `test_grad_edit_with_multiple_invokers` — assigning `x.grad = x.grad * 3` inside a batched invoke is applied.

```
pytest tests/test_lm.py::TestGradients tests/test_lm.py::TestInvokerBatching --device cpu
# 10 passed
```

All existing gradient and invoker-batching tests continue to pass (no regression).

### DIFF
--- a/src/nnsight/intervention/batching.py
+++ b/src/nnsight/intervention/batching.py
@@ -272,7 +272,13 @@ class Batcher:
         batch_start, batch_size = batch_group
 
         if acts.shape[0] == self.total_batch_size:
-            return acts.narrow(0, batch_start, batch_size)
+            view = acts.narrow(0, batch_start, batch_size)
+            # Tag the view so a backward hook registered on it can be redirected
+            # to the full-batch parent tensor, which is the tensor actually in
+            # the loss graph. Without this, gradients for batched invokes are
+            # never computed for the view (see wrap_grad in tracing/backwards.py).
+            view._nnsight_batch_group = (batch_start, batch_size)
+            return view
 
         return acts
 
--- a/src/nnsight/intervention/tracing/backwards.py
+++ b/src/nnsight/intervention/tracing/backwards.py
@@ -17,9 +17,15 @@ def wrap_grad(interleaver: Interleaver):
 
     def wrap(tensor: torch.Tensor):
 
-        # Only wrap the tensor once
-        if tensor._backward_hooks:
-            return
+        # When two or more invokes share an input, the Batcher delivers each
+        # invoke a storage-sharing view (`tensor.narrow(0, start, size)`) of the
+        # full-batch activation. That view is never an input to any op producing
+        # the loss, so a hook on it never fires. The Batcher tags such views with
+        # their batch slice; redirect the hook to the full-batch parent and
+        # narrow the gradient to this invoke's slice. The parent IS in the loss
+        # graph, so its hook fires. (See Batcher._narrow in ../batching.py.)
+        batch_group = getattr(tensor, "_nnsight_batch_group", None)
+        redirect = batch_group is not None and tensor._base is not None
 
         # We are providing the grad of the tensor
         provider = id(tensor)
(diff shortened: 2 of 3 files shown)


JadenFiotto-Kaufman:
Thanks for this, @azrabano23 — and sorry for the slow turnaround! Merged. 🚀

Cross-version note, since we're mid-migration to the pipeline-based **0.8**: the `tracing/backwards.py` machinery this patches was removed in the rewrite, so the PR doesn't port directly — but the **bug reproduces in 0.8** (a `.grad` read inside a batched invoke hangs, surfacing as `OutOfOrderError`, 0.8's flavor of the old `MissedProviderError`). Your fix inspired the 0.8-native equivalent:

https://github.com/ndif-team/nnsight/commit/a21d6a2cc20bd275317e7543b0a0d1716ca2eedb

A couple of details that lined up nicely with your analysis:

- Same root cause and same core technique — redirect the grad hook to the view's `._base` and recover the invoke's rows with **`as_strided(view.shape, view.stride(), view.storage_offset())`**. Your point about needing `as_strided` (not a dim-0 `narrow`) because the base is the **flattened `(batch·seq, hidden)`** activation holds in 0.8 too — I confirmed the narrowed outputs are `.view()`s of that flattened base, so the strided recovery is exactly what's required.
- One addition: 0.8 tags the batcher-created slice with a small marker and gates the redirect on it, so **user-made views keep the direct-hook behavior** — sidestepping the ambiguity you flagged in the issue's open question. (The tag is a bare bool, not the parent tensor, to keep saved activations cheap to serialize for remote.)

Regression tests for both the read and edit paths are in the commit. Thanks again 🙏

PR #679, opened 2026-06-25
refactor(interleaver): run interventions as greenlets, not OS threads

### LINKED ISSUE #1: Maybe make Optimization a torch module.

JadenFiotto-Kaufman:
## Summary

Replaces the per-invoke **OS-thread** worker model in the interleaver with **greenlet** stackful coroutines. The intervention engine is serial-by-design (only one thread of control is ever runnable — the worker ping-pongs with the model), so an OS thread per invoke was paying for concurrency that never existed. greenlet's `switch` replaces the lock-based queue handoff **1:1**: a worker yields control with its event and resumes with the response, all on one OS thread.

This deletes machinery rather than adding it: the lock-based single-slot queue, the response queue, the per-worker CUDA-stream propagation, and the thread lifecycle all go away.

## Results

- ✅ **Whole local suite green: 288 passed, 2 skipped, 1 xpassed** — basic tracing, gradients/backward, multi-invoke, barriers, iteration, early-stop, `.source`, caching, `model.edit()` with attached modules, serialization, VLM.
- ⚡ **~1.6× faster per trace** (threads 389 µs → greenlet 245 µs on a single-access micro-benchmark). The per-invoke thread spawn + lock handoff was the overhead removed.
- Net `interleaver.py`: +87 / −65, mostly deletions.

## What changed

**`interleaver.py`**
- `Mediator.Value` (a lock-as-Event single-slot queue) → `Mediator.Pending`, a **lockless** one-slot holder. The greenlet `switch` provides the synchronization.
- `response_queue` **deleted** — the response to an event is simply the return value of switching back into the worker.
- `self.worker`: `threading.Thread` → `greenlet`. Worker yields via `self._resumer.switch((event, data))`; main resumes via `self._resume(value)`.
- **CUDA-stream capture/propagation deleted** — the worker now shares the caller's OS thread (and therefore its stream), so the race that propagation guarded against cannot occur.
- `cancel()` simplified: resume a stuck worker with `Cancelation()` instead of the manual `put/get` drain.

**`tracing/backwards.py`**
- Backward runs under `torch.autograd.set_multithreading_enabled(False)` (see fix #1).

**`pyproject.toml`**: added `greenlet>=3.0.0`.

## Two non-obvious fixes (both genuine, not workarounds)

1. **Backward must run single-threaded.** PyTorch runs the CUDA backward on a dedicated device thread, so grad hooks fire on a *different* thread than the one that created the worker greenlet → `greenlet: cannot switch to a different thread`. Wrapping the backward in `torch.autograd.set_multithreading_enabled(False)` moves grad hooks onto the calling thread (verified empirically). This also aligns with nnsight's single-threaded-by-design interleaver.
2. **Dynamic `_resumer`, not a fixed "main".** A worker can resume a *sibling* mediator mid-body — e.g. `model.edit()` whose body calls an attached, hooked module. That sibling must return control to *whoever resumed it*, so `_resumer` is refreshed on every resume rather than captured once in `start()`. The thread model got this for free because the queue handoff is thread-agnostic; greenlet makes the control-flow topology explicit. Caught by the editing tests (`cannot unpack MissedProviderError`).

## Known limitation / follow-up

⚠️ **greenlet cannot switch across OS threads.** Any path where the model forward (and thus hooks) runs on a different thread than mediator creation will hit "cannot switch to a different thread." Fixed for backward (above). **vLLM and remote/serve paths are not yet validated** (no GPU-vLLM / NDIF infra in the dev environment). The interleaver appears to be created and entered within `execute_model` (same thread) and the async backend uses a single asyncio event loop, so it is *likely* fine — but **must be run against `test_vllm.py` / `test_remote.py` / `test_serve*.py` before merge.** If a cross-thread case surfaces, the same pattern as fix #1 applies: ensure the worker greenlet is created on the thread that drives it.

See `notes-greenlet-conversion.md` for full details.


JadenFiotto-Kaufman:
Superseded by https://github.com/ndif-team/nnsight/pull/680 — this branch was accidentally cut from `hackathon/peft` and carried 13 unrelated peft commits. The replacement is branched cleanly off `dev` with only the greenlet changes.

PR #681, opened 2026-07-08
Fix cache navigation through collapsing/re-mount renames

Butanium:
## Summary

Fixes cache attribute/index navigation when using collapsing or re-mount renames (e.g., `rename={"model.layers": "layers"}`). These renames re-mount nested modules at the root, creating alias paths that don't exist in real storage keys. The cache navigator now resolves these paths correctly through an alias-path map.

## Key Changes

- **Added `_resolved_path()` method**: Translates aliased paths (from collapsing renames) back to their real storage keys using the `_alias_paths` map.

- **Updated property accessors** (`.output`, `.inputs`, `.input`): Now use `_resolved_path()` to look up cache entries, enabling access through renamed paths.

- **Added `_alias_path_prefix()` helper**: Checks if a path is an alias path or a prefix of one, allowing segment-by-segment navigation through the alias-path keyspace.

- **Refactored `_add_alias_path()`**: Protects the root cache component from substring replacements. A rename like `{"model": "foo"}` now correctly produces `model.foo.layers.0` (not `foo.foo.layers.0`), since `cache.model` is the fixed root.

- **Enhanced `__getitem__()` (integer indexing)**: Now checks both real keys and alias-path prefixes when validating numeric indices, raising `IndexError` instead of leaking `KeyError` for out-of-bounds access on renamed module lists.

- **Refactored `__getattr__()` (attribute access)**: Implements a three-route resolution strategy:
  1. Direct route: path prefixes a real storage key
  2. Leaf-alias route: translate aliased segment and continue in real-path space
  3. Alias-path route: path prefixes an authoritative alias path (handles collapsing renames)

- **Added `_child()` helper**: Centralizes sub-view creation to reduce duplication.

## Tests

Added two new test cases:
- `test_cache_collapsing_rename`: Validates attribute/index access through collapsing renames, including `.output`, `.input`, and out-of-bounds indexing.
- `test_cache_rename_preserves_root`: Ensures renames matching the root component name don't corrupt the fixed `cache.model` root.

## Implementation Details

The core issue: when a rename like `{"model.layers": "layers"}` is applied, the user-facing access path `cache.model.layers[0]` doesn't match any real storage key (which is `model.transformer.h.0`). The fix maintains an `_alias_paths` map that records these mappings and uses it during navigation to resolve paths correctly. The three-route resolution in `__getattr__()` ensures both simple renames and complex collapsing renames work seamlessly.

https://claude.ai/code/session_01CGecSNRudbfxSX14gsBco5

### DIFF
--- a/src/nnsight/intervention/tracing/tracer.py
+++ b/src/nnsight/intervention/tracing/tracer.py
@@ -99,30 +99,45 @@ class Cache:
             # (e.g. translating root key ``"model"`` → ``"transformer"``,
             # which is not in storage).
             super().__init__()
-            if data:
+            # Use the raw dict length, not the ``_path``-scoped ``__len__``:
+            # an alias-space sub-view (from a collapsing/re-mount rename) has a
+            # scoped length of 0 because no real key lives under its alias path,
+            # yet its full storage must still propagate to deeper children.
+            if dict.__len__(data):
                 for k in dict.__iter__(data):
                     dict.__setitem__(self, k, dict.__getitem__(data, k))
+
+        def _resolved_path(self):
+            """Resolve ``_path`` to a real storage key.
+
+            When navigation lands on an aliased path produced by a
+            collapsing/re-mount rename (e.g. ``"model.layers.0"`` from
+            ``rename={"model.layers": "layers"}``), the underlying storage key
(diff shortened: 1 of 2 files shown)


Butanium:
#682 for fable less sloppy version

PR #682, opened 2026-07-08
Robustify activation caching: rename navigation, hook lifecycle, membership semantics

### LINKED ISSUE #535: Unused renaming scheme make cache access via renamed name fail

Butanium:
## Summary

Started as a fix for #535 (cache attribute access broken under collapsing/re-mount renames) and grew into a robustness pass over the activation-cache subsystem. Three headline changes:

1. **Cache navigation is structural and shares rename semantics with the envoy tree** — fixes #535 and the whole class of string-matching bugs around it.
2. **Cache hooks are actually removed at trace exit** — previously they leaked, and the *next* trace on the same model silently corrupted the old cache.
3. **Membership (`in`) and lookup (`[]`) share one resolution semantics** — renamed spellings included.

## 1. Rename navigation (fixes #535)

Cache storage keys are always **real** module paths (dict-style access, `keys()`, iteration unchanged). Navigation resolves against a **weights-free skeleton of the envoy tree**:

- `Envoy._skeleton()` snapshots the tree as `PathNode`s: real path, children by name, aliases resolved to their *target nodes* via `Aliaser.resolve(envoy, alias)` — the **same function** `Envoy.__getattr__` uses, so cache and envoy cannot disagree about what an alias means. Nodes hold no modules: the skeleton pickles, travels with remote results, and keeps no weights alive.
- `CacheDict` navigation is per-segment tree descent (`PathNode.resolve`, real children authoritative over aliases). Navigation refuses to descend into subtrees where nothing was cached, and error messages now distinguish *"exists on the model but was never cached"* from *"no such module"*.
- Every alias of a module navigates — `cache.model.transformer.h[x]`, `cache.model.model.layers[x]`, `cache.model.layers[x]`, and secondary aliases of collapsing renames (`{"model.layers": ["layers", "blocks"]}`) all reach the same entry. `test_cache_navigation_matches_envoy` sweeps the spellings and asserts envoy/cache equivalence.
- List-valued renames (documented Aliaser syntax) no longer crash `tracer.cache()`; dot-prefixed alias values (`{".transformer.h": ".layers"}`, also documented) now work on the envoy side too (`Aliaser.build` normalizes them).
- Out-of-bounds indices on (renamed) module lists raise `IndexError`; index 1 can no longer string-prefix-match `h.10`. Old pickles (no skeleton fields) fall back to a storage-derived trie, as do `CacheDict`s constructed without a model.
- Also fixed along the way: shared-mutable-default state across all `CacheDict`s in a process; the fixed `cache.model` root being corrupted by renames like `{"model": "foo"}`.

## 2. Hook lifecycle

Persistent cache hooks were never removed: dead mediators are pruned from `interleaver.mediators` (in `__exit__` and `check_dangling_mediators`) *before* `cancel()` runs `remove_hooks()`, so the handles leaked onto the modules. Consequence: run a second `model.trace(...)` on the same model and the first trace's cache hooks fire again — entries silently become `[Entry, Entry]` lists and previously-working `.output` calls raise `AttributeError: 'list' object has no attribute 'output'`.

Hooks can't be removed the moment a mediator dies (vLLM re-enters the interleaver context per engine step and cache hooks must keep recording across steps), so pruned mediators' handles are **retired** to the interleaver and drained in `cancel()` — the documented single per-trace cleanup path. `test_cache_hooks_removed_after_trace` asserts hook counts return to baseline and a prior cache survives subsequent traces intact.

## 3. Membership = lookup

`"model.layers.0" in cache` was `False` while `cache["model.layers.0"]` succeeded. `__contains__` now delegates to `__getitem__`, so `in` covers renamed spellings and sub-view-relative keys exactly like access does.

## Tests

Nine new/extended tests in `tests/test_lm.py::TestCache` plus one in `TestRename` (dotted alias values): collapsing renames (incl. list-valued and secondary aliases), root protection (issue #535's original example), envoy/cache navigation equivalence sweep, membership/lookup agreement, state isolation, alias-never-shadows-real-key, old-pickle compatibility, and hook removal across traces.

`tests/test_lm.py`: **90 passed** · `test_tiny.py`: 18 · `test_envoys/test_source/test_iter_edge_cases/test_memory_cleanup/test_multiple_wrappers/test_transform/test_0516_features`: 115 passed, 1 xpassed (all on `dev` base). All three repro snippets from the #535 thread verified. The branch was also adversarially reviewed by an independent agent (A/B behavioral comparison against `dev` on seven access patterns, pickle/deepcopy round-trips, multi-fire generation, swap renames); its findings are addressed in the later commits.

https://claude.ai/code/session_01GPKVtGmUQWFsS7dMEUsNNe

### DIFF
--- a/docs/usage/rename-modules.md
+++ b/docs/usage/rename-modules.md
@@ -106,11 +106,11 @@ model._update_alias({"new_name": "transformer.h"})
 
 Cache storage keys are always the **real** module paths, but navigation
 understands the rename. At cache creation, `tracer.cache(...)`
-(`src/nnsight/intervention/tracing/tracer.py`) asks the envoy tree for every
-module's user-facing path (`Envoy._aliased_paths`,
-`src/nnsight/intervention/envoy.py`) — the same `Aliaser` state that resolves
-`model.layers` — and `Cache.CacheDict` uses that map to translate renamed
-access paths back to real storage keys:
+(`src/nnsight/intervention/tracing/tracer.py`) snapshots the envoy tree as a
+weights-free skeleton (`Envoy._skeleton`,
+`src/nnsight/intervention/envoy.py`) whose nodes carry children and aliases —
+resolved from the same `Aliaser` state that answers `model.layers` — and
+`Cache.CacheDict` navigates that skeleton to reach real storage keys:
 
 ```python
 with model.trace("Hello") as tracer:
--- a/docs/usage/rename-modules.md
+++ b/docs/usage/rename-modules.md
@@ -104,15 +104,25 @@ model._update_alias({"new_name": "transformer.h"})
 
 ## Cache keys honor the rename
 
-`tracer.cache(...)` (`src/nnsight/intervention/tracing/tracer.py:465`) passes `rename` and the inverted alias map into `Cache.CacheDict`, so:
+Cache storage keys are always the **real** module paths, but navigation
+understands the rename. At cache creation, `tracer.cache(...)`
+(`src/nnsight/intervention/tracing/tracer.py`) asks the envoy tree for every
+module's user-facing path (`Envoy._aliased_paths`,
+`src/nnsight/intervention/envoy.py`) — the same `Aliaser` state that resolves
+`model.layers` — and `Cache.CacheDict` uses that map to translate renamed
+access paths back to real storage keys:
 
 ```python
 with model.trace("Hello") as tracer:
     cache = tracer.cache()
 
-cache["model.layers.0"].output    # works via alias
-cache.layers[0].output            # also works
+cache["model.transformer.h.0"].output   # real storage key
+cache["model.layers.0"].output          # renamed key, translated for you
+cache.model.layers[0].output            # attribute-style through the rename
 ```
+
+Real keys are authoritative: if a renamed path collides with a genuinely
+cached module's real path, the real module wins.
 
 ## Gotchas
 
--- a/src/nnsight/intervention/envoy.py
+++ b/src/nnsight/intervention/envoy.py
@@ -751,16 +751,16 @@ class Envoy(Batchable):
         root = build(self)
 
         # Second pass, once every node exists: link aliases to their target
-        # nodes. ``fetch_attr`` resolves dotted names through the aliasers
-        # themselves, so chained renames compose exactly as attribute access
-        # does.
+        # nodes through ``Aliaser.resolve`` — the same call envoy attribute
+        # access makes, so cache navigation cannot disagree with it (and
+        # chained renames compose exactly as attribute access does).
         def link(envoy: Envoy):
             aliaser = envoy._alias
             if aliaser is not None:
                 node = nodes[envoy.path]
-                for alias, name in aliaser.alias_to_name.items():
+                for alias in aliaser.alias_to_name:
                     try:
-                        target = util.fetch_attr(envoy, name)
+                        target = aliaser.resolve(envoy, alias)
                     except Exception:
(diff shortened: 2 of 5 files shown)


JadenFiotto-Kaufman:
Closing this as superseded. The rewrite reached the same conclusions from a
different direction, and this predates it by about two weeks.

Two of the five files no longer exist on `0.8`:
`src/nnsight/intervention/tracing/tracer.py`, which is the +219/-88 that makes up
most of the PR, and `tests/test_lm.py`, which is all of its tests. So there is
nothing to port directly. On each of the three fixes:

Cache navigation under renames, issue #535. Resolved the way you proposed.
Navigation on `0.8` resolves against the envoy tree instead of matching strings,
so an alias means the same thing to the cache and the envoy. Hotragn checked #535
against `0.8` recently while working on #701 and found it mostly does not
reproduce.

Cache hooks leaking at trace exit. Moot on `0.8`, where `cache.py` has no hook
references at all. The rewrite replaced persistent forward hooks with a
per-module controller, so there are no handles to leak and no second trace that
can re-fire the first one's.

`in` disagreeing with `[]`. Already the case: `CacheDict.__contains__` delegates
to `__getitem__` and returns False on the lookup errors, at
`intervention/cache.py:313`, so membership covers renamed spellings the same way
access does.

Worth saying what was good here anyway. The hook leak, where a second trace
silently corrupts the first trace's cache, was a real bug and a nasty one.
Distinguishing "exists on the model but was never cached" from "no such module"
is the right error design. The shared mutable default state across every
`CacheDict` in a process was a good catch.

Please do re-check `0.8` when you get a chance. You reported #535 and you know
this subsystem better than the rewrite's test coverage does. If any of these
still reproduce there, or if the new navigation gets an alias case wrong that
yours handled, open an issue and I will take it seriously.

One process note, and it is our fault rather than yours: `CLAUDE.md` still says
to base PRs on `dev` while all development happens on `0.8`. Two open PRs have
been misdirected by that. We will fix the guidance.

PR #686, opened 2026-07-23
0.8 — the pipeline rewrite

JadenFiotto-Kaufman:
## TL;DR

**0.8 is a ground-up rewrite of nnsight's execution engine.** The mental model is unchanged — open `with model.trace(...)`, then read, edit, and `.save()` a model's internals as ordinary Python — but the core underneath is new: intervention code and the forward pass now **interleave as greenlets** instead of coordinating across OS threads.

The public API is largely familiar. A handful of things moved, were deprecated, or now raise where they used to no-op — the migration notes are at the bottom.

> **This description was rewritten on 2026-09-08.** The original was written at branch time and had gone stale in three ways: the LOC figures predated the model work that landed since, the performance tables were measured before the `perf/interleaver` commits, and tensor parallelism, quantization, the vLLM rebuild and the `.source` overhaul weren't in it at all.

---

## By the numbers (`main` 0.7 → `0.8`)

| | 0.7 (`main`) | 0.8 |
|---|---|---|
| Diff vs `dev` | — | **433 files, +51.5k / −61.8k**, 174 non-merge commits |
| Execution model | thread workers + locks/queues | **greenlets** (cooperative, single-threaded) |
| Engine (`intervention/`) | 7,336 code lines | **4,911** (−33%) |
| Model support (`modeling/`) | 5,126 code lines | **6,021** (+17%) |
| Whole package | 13,594 lines / 78 modules | 12,466 / 72 |
| Docs | — | 105 recipe pages, every claim executed |

The shape of that is the point: the engine got a third smaller while the surface it drives got bigger. `modeling/` grew because tensor parallelism, quantization, the diffusion runtime and the vLLM rebuild all landed on this branch.

> **Base branch:** this targets `dev` (it was cut from `dev`, so the diff is exactly the rewrite). Heads-up: `main` has commits not in `dev`/`0.8` that should be reconciled separately before a `dev → main` release.

---

### Execution: greenlets, not threads
Each intervention block runs in its own worker greenlet (a **`Mediator`**) that cooperatively hands control back to the model side whenever it parks on a location. Only one greenlet runs at a time, so there are **no locks or queues**. One shared **`Interleaver`** installs persistent pass-through forward hooks on every module; the worker/model event protocol is `VALUE` / `SWAP` / `SKIP` / `BARRIER`.

Reading a location the model already ran past raises **`OutOfOrderError`**, naming the location and pointing the traceback at the line that was waiting. (0.7 raised here too, but as `MissedProviderError: ... was not provided. Did you call an Envoy out of order? Investigate why this module was not called.` — which sent people hunting for a module-path typo. The gain is the diagnosis, not the absence of a hang.)

### Model classes
- **`NNsight(module)`** — base wrapper for any `torch.nn.Module`.
- **`TransformersModel("repo/id", task=...)`** — the **primary** HuggingFace class, backed by a `transformers.pipeline`; supports any task (text, vision, multimodal, audio), chunked tasks as rows of one forward, pre-loaded module instances, `peft=` adapters, and `rename=` / `envoys=` customization.
- **`DiffusionModel`** — any `diffusers` pipeline, UNet- **or** transformer-based (SD/SDXL, Flux/SD3/DiT), with `seed=`, per-invoke batching, denoising-step iteration, and `automodel=` to choose the class the weights load through.
- **`VLLM(..., mode="sync"|"async")`** — vLLM-backed runtime with interventions inside the engine worker (GPU; excluded from CPU CI).
- **`LanguageModel` / `VisionLanguageModel`** — now **deprecated** thin subclasses that warn on construction; use `TransformersModel(task=...)`.

### Tensor parallelism (landed after the original description)
`TransformersModel(..., distributed_config=DistributedConfig(tp_size=N))` under `torchrun`, ported to the **transformers 5.16 DTensor backend**. Sharded activations are gathered before an intervention sees them and re-split afterwards, so the trace reads and writes whole tensors exactly as it would on one GPU. 5.16 moved the sharding plan off the individual modules onto a model-level glob-pattern `_tp_plan`, moved the parallel style's transforms from forward hooks into a `forward` replacement (the slot nnsight's controller uses), and removed the `all_gather` / `split` helpers; nnsight reads the model-level plan and installs its controller around the style's wrapper. On 5.15 or earlier nothing is recognized as sharded and a trace sees one rank's slice.

### Quantization (landed after the original description)
You name a quantization where you would name a dtype: `TransformersModel(..., dtype="nf4")`. `nf4`/`int4`/`4bit` mean NF4; `fp4`, `int8`, `fp8` mean what they say. Module paths and activations are unchanged. `fp8` is **refused** below compute capability 8.9, because transformers otherwise logs a warning and silently loads bfloat16 at twice the requested width.

### vLLM (substantially rebuilt after the original description)
CUDA-graph taps, so interventions survive graph replay on the decode hot path. Engine-wide **`model.edit()`** installs a block once on the engine, and every later request gets its own scoped copy — including requests from clients that never heard of nnsight, which is what makes `nnsight-serve` work. Named edits selectable per request via `edits=[...]`, `n>1`, MoE fragment handling, prefix-cache recompute so cached tokens still reach interventions, and async streaming.

### `generate` vs `pipe`
`model.generate(...)` generates through the **model** and returns **token ids** on `tracer.result`, decoding with the checkpoint's own `generation_config`. `model.pipe(...)` runs the whole task **pipeline** and returns its records (decoded text, labels) — what the old `generate` returned. `model.trace(...)` runs one forward; `model.scan(...)` runs one forward under fake tensors for shape inference.

### Source tracing
`module.source.<op>` hooks intermediate operations inside a `forward` (AST-rewritten, lazy, inert outside a trace, recursively drillable). Since the original description: **every assignment is an operation** on the same per-name counter as calls, and **decorated forwards no longer raise `SourceNotAvailable`** — a wrapper is peeled and rebuilt around the instrumented function, and closures and `super()` forwards are instrumented too. `SourceNotAvailable` now means only "no Python source" (a builtin or C function). See the migration note about shifted labels below.

### `eproperty` — the hookable-value descriptor
`.input` / `.inputs` / `.output`, `tracer.result`, and `VLLM.logits` / `.samples` are all built on the reintroduced `@eproperty` descriptor (`preprocess` / `.postprocess` / `.transform` / `.provide`, with `description=` surfacing a value in the Envoy repr tree). You can declare your own hookable values on a model subclass.

### `save` is guarded
`nnsight.save(x)` / `x.save()` now **raises** if called outside a trace (it used to silently no-op). Idiom for collecting per-step values: save the container once, append raw values (`xs = nnsight.save([]); xs.append(...)`) — appending `x.save()` drops values on remote.

### Iteration
Target occurrences with `for step in tracer.iter[...]:` or `tracer.all()`. A **bounded** `iter[:N]` that outruns the run is cut short with a warning: values saved inside the loop are kept, statements after it are discarded. The `with tracer.iter[...]:` form is deprecated; `tracer.next()` is gone.

### Cache, scan, skip, stop, batching
`tracer.cache(...)` records many modules at once (synchronously by default, with a `non_blocking` flag for the device transfer); `module.skip(replacement)` bypasses a module; `tracer.stop()` exits early; `model.scan(...)` infers shapes under fake tensors. `tracer.invoke(prompt)` batches multiple prompts into one forward with per-invoke row narrowing, and refuses a write that would change an invoke's row count.

### Remote (NDIF)
- `remote=True` → `RemoteBackend` (blocking over one websocket, or `blocking=False` submit/poll).
- `remote="local"` → in-process serialize/deserialize/execute dry run — catch serialization issues offline, no key.
- `AsyncRemoteBackend` — `await` a job for its saves, `async for` streamed status.
- `model.session(remote=True)` bundles multiple traces into one job. Model identity via `to_model_key()` / `from_model_key()`.

### Config, packaging, tooling
`CONFIG.API.HOST` / `APIKEY` / `COMPRESS`, `CONFIG.APP.DEBUG` / `REMOTE_LOGGING` / `PYMOUNT`, loaded from `~/.config/nnsight/config.yaml` over shipped defaults, then env. `pip install nnsight` is a working local **+** remote install, with `dev` / `vllm` / `serve` extras; `nnsight` and `nnsight-serve` console entry points; `nnsight.login()` and `nnsight login`; **Python 3.10 through 3.14**, both ends tested in CI; PyPI publishing via Trusted Publishing.

---

## Docs & tooling
- **`NNsight.md`** — a design-and-implementation manual.
- **`docs/`** — 105 recipe-style pages (usage / concepts / patterns / errors / gotchas / remote / models / developing), routed for humans and agents by **`CLAUDE.md`**. Every claim was checked by executing it, which is where a good fraction of this branch's later fixes came from.
- **`NNsight_Walkthrough.ipynb`** — a runnable, Colab-ready guided tour.
- **CI** — installs `.[dev]`, runs the CPU suite (vLLM excluded), and runs the tensor-parallel tests on CPU over gloo. Agent support via Context7 and the nnsight skill.

---

## Testing

`CUDA_VISIBLE_DEVICES="" pytest tests/ --ignore=tests/vllm` collects **1,146 tests**. The vLLM suite (`tests/vllm/`) needs a CUDA GPU + `nnsight[vllm]`.

**A pass count needs a quiet machine, and should be measured before merge.** Two consecutive runs of the identical command on the same commit gave `16 failed, 1031 passed` and `101 failed, 924 passed`, on a box with other pytest runs competing for CPU. `tests/test_source.py` and `tests/test_tracing.py` account for most of the difference and **pass 132/132 in isolation**, so the variance is either load-related or cross-test state leakage worth a look on its own — source instrumentation persists for the life of the process and `BLOCKS` memoizes compiled blocks per call site.

Separately, `tests/tp/` and `tests/test_tensor_parallel_rules.py::TestStyleCoverage` **error rather than skip** below the transformers 5.16 floor (`ModuleNotFoundError: No module named 'transformers.distributed.tensor_parallel'`). The test's own comment says the failing import "is the point", but as written it turns CI red for anyone on 5.15.

---

### Per-intervention overhead
Measured on a compute-free stack of 96 tiny blocks, CPU, pinned to isolated cores, with **the same model's untraced forward subtracted out** so the model's own compute cancels; median of two runs per version, and the two versions' baseline forwards agreed to within 2%.

| | 0.7 (`main`) | 0.8 |
|---|---:|---:|
| Opening a trace with nothing in it | 1.12 ms | **0.51 ms** |
| Each activation read | 36.7 µs | **11.8 µs** (3.1× flatter) |
| Each activation edited | 43.9 µs | **17.1 µs** (2.6× flatter) |
| Each extra prompt batched into one pass | 94.3 µs | **15.6 µs** (6.0× flatter) |

At the ends of the sweeps: 192 activations read in one pass costs 8.2 ms on 0.7 and 2.8 ms on 0.8; 128 prompts batched into one forward costs 13.1 ms on 0.7 and 2.6 ms on 0.8.

The batching column is where the rewrite really shows — 0.7's `Interleaver.handle` fanned every visit out to every mediator, so cost scaled with modules × invokes.

### End to end on a real model
Not reported. GPT-2-on-CPU numbers were attempted and are not trustworthy on shared hardware: the same untraced forward measured anywhere from 37 ms to 52 ms across runs. The arithmetic is the honest version — a dozen reads at 11.8 µs is 0.14 ms against a ~50 ms forward, about a quarter of one percent — so for a few interventions on a large model this is a wash, and the greenlet engine pays off when a trace touches a lot of the model at once.

Repro: `tests/performance/interleave_bench.py` (version-agnostic; point `PYTHONPATH` at each tree and diff with `compare.py`). The sweeps above extend that harness to 192 reads/writes and 128 invokes and subtract the baseline forward.

---

## Known follow-ups (not in this PR)
- **Streaming/hybrid remote** (`tracer.local()` — run a function back on the client) is not ported.
- **Serialization edges:** frame objects don't serialize; sourceless functions fall back to cloudpickle **silently** (no `__source__` escape hatch); file-defined remote functions using `nonlocal` fail at load.
- **Multimodal *generate* batching** raises `NotImplementedError` (single multimodal generate works).
- **Dropped extensibility/ergonomics:** `.pyi` stubs, arbitrary-method trace wrapping.
- **Security (greenfield, absent in 0.7 too):** no deserialization allow/deny layer — a deliberate hardening decision for remote servers.
- **TP tests error instead of skipping** below transformers 5.16 (above).

---

## Migration quick reference

Two changes can alter results **without raising anything**:

- **`.source` operation labels shifted.** Every assignment is now an operation sharing the per-name counter with calls, so GPT-2's attention call moved `attention_interface_0` → `attention_interface_1` (`attention_interface_0` is now the line that chooses the implementation). Requesting the old label doesn't raise — it returns the assigned value instead.
- **A bounded `tracer.iter[:N]` that outruns the run is cut short** with a warning, keeping saves and discarding trailing statements, so a result can look complete while being shorter than the bound. 0.7 bounded `all()` internally; code written against it should hold the run to the count (`min_new_tokens=`, or `min_tokens=` / `ignore_eos=True` on vLLM) or move trailing statements into a separate `tracer.invoke()`.

The rest raise:

| 0.7 | 0.8 |
|---|---|
| `LanguageModel(...)` / `VisionLanguageModel(...)` | `TransformersModel(repo, task=...)` — old names still work, but warn |
| `model.generator.output` (generated ids) | `tracer.result`; `model.pipe(...)` for decoded pipeline records |
| `x.save()` / `nnsight.save(x)` outside a trace (silent no-op) | **raises** — call it inside the trace |
| `saved.value` | a saved value *is* the value |
| `tracer.next()` / `module.next()` / `with tracer.iter[...]:` | `for step in tracer.iter[...]:` / `tracer.all()` |
| `nnsight.apply/log/cond/iter/session(...)`, `nnsight.list/dict/int/...` | plain Python + `model.session()` |
| `CONFIG.APP.CROSS_INVOKER` / `CACHE_DIR` / `TRACE_CACHING` | gone |
| custom `_prepare_input` / `_batch` | new signatures |

Everything reachable under an old name warns under `nnsight.NNsightDeprecationWarning`, a `FutureWarning` (not a `DeprecationWarning`, whose default filters hide it anywhere but the top level of the running script).

See [`docs/reference/version-history.md`](docs/reference/version-history.md) for the full mapping.


AdamBelfki3:
🙂

khaiwang:
# 0.8 review

I ran the 0.8 branch on real GPUs and against an interpretability-workload
benchmark (56 cells, HF and vLLM-sync backends, semantics compared against an
HF control). Everything below was reproduced on this machine: py3.11, torch
2.11, transformers 5.12.1 for local runs; vllm 0.19.1 with transformers 4.57
for engine runs; 2×A100 for the GPU suites. Repro commands are in each item.

## Fixes

**Worker-side VLLM(module) construction.** At the PR head, every vLLM GPU
worker dies at startup: the runner wraps the already-loaded module with
`VLLM(self.model)`, `_load` assumes a repo-id string, and the module lands in
`EngineArgs` (`'GPT2LMHeadModel' object has no attribute 'lower'`; repro:
`tests/vllm/test_tracing.py`, TP=1). The PR's CI covers the CPU suite only,
so the vLLM path was never executed on a GPU. Fix: a ready module wraps
directly, and the `interleaver=` kwarg threads to the Envoy the way
`rename`/`envoys` already do instead of leaking into the load path.

**Barrier exception attribution.** Two blocks meet at a `tracer.barrier(2)`;
the block that resumes after release raises. The exception surfaces inside
the releasing block's greenlet, unwinds it mid-body, and gets recorded on the
releasing mediator: with `defer_exceptions` (the engine driver mode) the
wrong invoke's request fails while the raising invoke returns truncated work
as success. Both blocks belong to one trace, so this is a diagnosability and
partial-result bug for batched traces; other traces are unaffected. Fix: the
release loop records the exception on the mediator that raised it, keeps
releasing the remaining waiters under a deferring interleaver, and re-raises
otherwise with the raiser's traceback already stashed.

**Batched saves on vLLM.** The documented batching idiom (a container saved
above the invoke blocks, each invoke writing its slot) returns the container
empty. Two causes. `.save()` marks object ids, which do not cross the
process boundary, so a serialized mediator arrives in the worker with no
record that its captured variables were saved. And each invoke rides its own
request with its own pickled copy of the container, so the copies come back
one per request and the last one clobbers the rest. Fix: `__getstate__`
ships the saved names with the mediator, and the sync collect merges
same-name copies element-wise before results push. The merge half is
interim: a per-trace worker scope (design question below) removes the copies
entirely, at which point `collect.py` deletes whole. The names half is
needed under any design, since saved-ness must always travel by name. Same
commit: the chat-input probe hard-imported `transformers.pipelines.base.Chat`,
which exists only from the transformers 5 chat-pipeline refactor, so every
text trace on a 4.x environment (what vllm 0.19 pins) died on ImportError;
the probe now feature-detects and treats input as not-chat when the helpers
are absent.

**Backward sessions on tensors outside autograd.** `with t.backward():` on a
tensor from an untracked forward fails today with
`RuntimeError: cannot register a hook on a tensor that doesn't require
gradient`, raised from the block's first `.grad` read. Inside an engine
worker it is worse: the shipped block's source is registered nowhere, the
nested capture fails, and the construct silently degrades to a bare
`t.backward()` followed by `TypeError: 'NoneType' object does not support
the context manager protocol`, with the bare backward having already run
outside the trace. Fix: `execute` refuses up front when the tensor does not
require grad (a requires-grad leaf still opens a session; autograd fires its
accumulation hook), and `Mediator.__setstate__` registers the deserialized
block's source in linecache under a unique filename, which 0.7's serializer
also did and which the rewrite dropped. Known limitation of the fix as
committed: the linecache entries are keyed by `id(self)` and never removed,
so a long-lived worker accumulates one entry per shipped block and an id
reuse can point an old traceback at the wrong source. A `weakref.finalize`
plus a monotonic counter closes both.

**Constructor argument routing.** Hardening on top of the minimal fixes: the
two decisions every model constructor makes (ready module vs load arguments,
Envoy kwargs vs load kwargs) each move to one place. `Loadable.__init__`
dispatches a ready module to `_wrap` and load arguments to `_load`, so no
subclass `_load` can receive a module again; the kwarg split derives from
`inspect.signature(Envoy.__init__)`, so the next Envoy parameter routes with
zero mixin edits. `tests/test_construction_routing.py` pins both invariants
across lazy construction and dispatch replay. Take it or leave it
independently of the fixes beneath it.

Validation: full CPU suite green on the branch (654 passed, 6 skipped),
single-rank GPU vLLM tracing green, and the benchmark's batched steering
cells match the HF control (top-1 agreement 1.00, total variation 0.000).

## Design question 1: a per-trace scope on engine workers

Locally, the invoke blocks of one trace share one namespace and one set of
objects through their frame. On the engine, each invoke's mediator is
pickled separately into its request's `extra_args`, so every shared object
arrives as a private copy per request. One missing link, four symptoms:

- shared containers come back empty or clobbered (patched by the interim
  merge above);
- a barrier never releases: each request holds its own copy, each copy waits
  for arrivals that go to the other copies, and every request retires with
  `ValueError: A barrier was never reached by every block it waits for`;
- a variable bound in one invoke and read in another resolves locally
  through the shared frame and is simply absent in the sibling's worker
  scope;
- any identity-based cross-invoke pattern silently operates on copies.

0.7 answered this server-side: requests carried a trace id, the first
arrival's globals became canonical, later arrivals were grafted onto them,
and the canonical values shipped home once when the trace completed. The
0.8 equivalent would be a per-trace scope keyed by trace id, established at
deserialization, with mediator scopes pointing at it.

The scheduler is the second half of the question. Continuous batching gives
no co-residency guarantee: with capacity for one request, invoke A runs and
retires (its parked barrier killed by the dangling check) before invoke B is
ever admitted, so shared state alone cannot make a rendezvous work. Either
a trace's requests are gang-scheduled (admitted together, none retired while
a sibling owes it a rendezvous), or the engine contract restricts invokes to
the schedule-independent subset: disjoint writes during the trace, merged
reads after it. Naming that contract is the decision; the interim merge
implements the restricted subset today.

## Design question 2: the gradient contract on inference engines

vLLM's forward runs under `torch.inference_mode` in both eras. On 0.7,
intervention code ran on its own OS threads outside the thread-local mode,
so torch itself raised on every gradient-surface touch: the 0.7 gaps report
records `requires_grad_()` failing with RuntimeError. Greenlets run
intervention code on the model's thread, inside the mode. That is what
makes in-place steering work on 0.8, and it also means `requires_grad_()`
now silently no-ops and gradient misuse fails deep inside the machinery
instead of at the door.

The backward refusal above covers one construct. The general decision: an
engine backend could declare centrally that it has no gradient surface, with
every entry point (`backward` sessions, `requires_grad_`, `.grad` access)
raising the same error, or it could grow a real gradient path by recomputing
the relevant forward segment outside inference mode. Until one of those is
chosen, a forward-only probe that flips `requires_grad` and expects
gradients later is silently wrong on 0.8 where 0.7 raised.

src/nnsight/modeling/mixins/loadable.py
JadenFiotto-Kaufman, line 1: 👍

src/nnsight/intervention/barrier.py
JadenFiotto-Kaufman, line 1: 👍

src/nnsight/intervention/interleaver.py
JadenFiotto-Kaufman, line 1: Hmm

src/nnsight/intervention/interleaver.py
```diff
@@ -236,8 +236,21 @@ def __getstate__(self) -> dict:
         return {"reduced": reduced, "copy": self.copy, "presaved": presaved}
 
     def __setstate__(self, state: dict) -> None:
```
JadenFiotto-Kaufman, line 269: Would be nice if we had some deduplication for this and `Request.deserialize` somewhere

Issue #695, opened 2026-08-24
Lint-time prohibited-module check is gone, its tests cannot import, and neither is in CI


Hotragn:
## Summary

The lint-time prohibited-module check added in ccb1702c (`feat(serialization): add security blacklist for prohibited modules`) is no longer present on `dev` or `main`. What is left behind is a comment saying it runs, a dead local variable, seven tests that cannot import, and no CI job that would have noticed.

I'm filing this rather than restoring it, because whether it should come back is a policy call — see the last section.

## What is missing

`PROHIBITED_MODULES`, `PROHIBITED_BUILTINS` and `PicklingProhibitedError` are all absent from `src/nnsight/intervention/serialization.py`:

```console
$ grep -rn "PROHIBITED\|PicklingProhibitedError" src/nnsight/
(no matches)
```

`ccb1702c` is an ancestor of `dev`, so this was lost somewhere in the 0.6/0.7 refactors rather than removed in a commit of its own — `git log -S PicklingProhibitedError` finds only the commit that added it and `bb3d09a4` on the `0.8` branch.

Verified on `dev` @ efe4244:

```python
>>> from nnsight.intervention.serialization import dumps
>>> import os
>>> def dangerous(): return os.getcwd()
>>> len(dumps(dangerous))     # serializes without complaint
```

## What is left behind

`serialization.py` (in `_dynamic_function_reduce`, ~line 826) still announces the check:

```python
# SECURITY CHECK: Check function globals for prohibited modules/functions.
# This catches dangerous patterns like `import os; os.getcwd()` or
# `from subprocess import run; run(...)` where the prohibited object
# is captured in the function's globals.
captured_globals = slotstate.get("__globals__", {})

# CLOSURE: Extract closure values and names.
```

`captured_globals` is assigned and never read anywhere in the file. To anyone auditing the code path that ships user functions to NDIF, this reads as though the globals are being validated here.

`docs/serialization_security.md`, added in the same commit, is also gone from `dev`.

## Why it went unnoticed

The seven tests that guarded it are still in the tree and still fail:

```
FAILED tests/test_serialization_edge_cases.py::test_prohibited_module_os
FAILED tests/test_serialization_edge_cases.py::test_prohibited_module_subprocess
FAILED tests/test_serialization_edge_cases.py::test_prohibited_module_socket
FAILED tests/test_serialization_edge_cases.py::test_prohibited_module_shutil
[...]

PR #696, opened 2026-08-24
test(serialization): make the serialization suite runnable and add it to CI

Hotragn:
## Summary

The four serialization test files are not in the CI list, and two of them had stopped working entirely without anyone noticing. This makes the suite runnable, removes the order-dependence in it, corrects a comment in `serialization.py` that claims a check which isn't there, and adds all four files to CI.

Found while checking which test files CI does *not* cover.

## What was broken

**1. `test_whitelist_serialization.py` did not collect at all.**

```
ImportError: cannot import name 'SERVER_MODULES_WHITELIST' from
'nnsight.intervention.serialization'
```

`SERVER_MODULES_WHITELIST` / `_is_whitelisted_module` were imported at module scope, so their removal took **all 12 tests** in the file down — including the nine that test source serialization and still pass. Seven tests in `test_serialization_edge_cases.py` failed the same way on `PicklingProhibitedError`.

Both are now resolved per-test and **skip with a stated reason** rather than erroring at import. The rest of each file runs, and these come back automatically if the APIs return — which felt better than deleting the maintainers' intent.

**2. Four tests were order-dependent and passed by accident.**

`test_no_register_needed`, `test_serialization_includes_module_functions`, and the two `*_serialized_by_source` tests all assert `b"def normalize" in dumps(normalize)`. Serializing by source is opt-in:

```
before register: 47 bytes, source embedded = False
after  register: 887 bytes, source embedded = True
```

`register` mutates a process-global set in cloudpickle (`_PICKLE_BY_VALUE_MODULES`), and every trace calls `get_local_env()`, which registers local modules. So these four passed in a whole-file run — and **every one of them failed when run alone**:

```
test_non_whitelisted_function_serialized_by_source  -> 1 failed
test_serialization_includes_module_functions        -> 1 failed
test_no_register_needed                             -> 1 failed
test_non_whitelisted_class_serialized_by_source     -> 1 failed
```

They now use a fixture that registers and unregisters around each test. A second fixture lifts the entries back out of the global set so the by-reference half of the contract can be asserted wherever in the file it runs. **Every test in the file now passes (or skips) in its own process.**

**3. `test_no_register_needed` asserted the opposite of current behaviour.**

Its premise — that `register()` is unnecessary — was true under the whitelist design that has since been removed. Without registration, `dumps` emits a GLOBAL the server cannot import, which is the `ModuleNotFoundError` that `docs/remote/register-local-modules.md` exists to prevent. Replaced with two tests that pin the real contract in both directions:

- `test_register_switches_to_serialization_by_value` — registration is what makes source travel with the job
- `test_unregistered_module_is_serialized_by_reference` — without it, the payload is a bare `GLOBAL mymethods.stateful normalize`

**4. `serialization.py` claimed a security check it no longer performs.**

```python
# SECURITY CHECK: Check function globals for prohibited modules/functions.
# This catches dangerous patterns like `import os; os.getcwd()` or
# `from subprocess import run; run(...)` where the prohibited object
# is captured in the function's globals.
captured_globals = slotstate.get("__globals__", {})

# CLOSURE: Extract closure values and names.
```

`captured_globals` is assigned and never read — it is the only remaining trace of the lint-time blacklist added in ccb1702c. Left as-is, this reads to anyone auditing the remote path as though input is being validated there. Replaced with a note stating the check is absent, where it would go, and that its tests are already written.

I have **not** restored the blacklist itself — see #695 for that question. It would start rejecting user code that references `pathlib`, `io`, `glob` and friends, which is a policy call rather than something to slip into a test-coverage PR.

## Result

```
before:  248 tests in CI
after:   323 passed, 12 skipped, 1 xpassed
```

75 previously-unrun tests, covering the path that ships user code to NDIF.

> [!NOTE]
> The CI run will stay red until #693 lands. `test_lm.py::TestGradients::test_backward_with_multiple_invokers` is already failing on `dev` and `main` and is unrelated to this change. Happy to rebase this on top of #693, or to drop the workflow change into a follow-up if you'd rather land the test repairs first.

<sub>Run with `transformers` 5.15.1 / `torch` 2.9.1 / `cloudpickle` 3.1.2 on CPU.</sub>

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -45,4 +45,8 @@ jobs:
           tests/test_mutual_recursion.py \
           tests/test_local_env.py \
           tests/test_local_simulation.py \
+          tests/test_dataclass_serialization.py \
+          tests/test_lambda_serialization.py \
+          tests/test_serialization_edge_cases.py \
+          tests/test_whitelist_serialization.py \
           --device cpu
--- a/src/nnsight/intervention/serialization.py
+++ b/src/nnsight/intervention/serialization.py
@@ -823,11 +823,18 @@ class CustomCloudPickler(cloudpickle.Pickler):
             if k in func_globals:
                 base_globals[k] = func_globals[k]
 
-        # SECURITY CHECK: Check function globals for prohibited modules/functions.
-        # This catches dangerous patterns like `import os; os.getcwd()` or
-        # `from subprocess import run; run(...)` where the prohibited object
-        # is captured in the function's globals.
-        captured_globals = slotstate.get("__globals__", {})
+        # NOTE: there is no prohibited-module check here. A lint-time blacklist
+        # (`PROHIBITED_MODULES` / `PROHIBITED_BUILTINS`, raising
+        # `PicklingProhibitedError`) used to run at this point and reject
+        # functions whose globals captured `os`, `subprocess`, `eval` and
+        # similar before they were serialized for NDIF. It is gone, along with
+        # the comment that used to describe it and a `captured_globals` local
+        # that nothing read -- which together left this reading as though the
+        # check were still here. Real sandboxing is server-side; if the
+        # client-side lint is wanted back it belongs on `slotstate["__globals__"]`
+        # right here, and the tests for it are already written (the
+        # `test_prohibited_*` cases in tests/test_serialization_edge_cases.py,
+        # currently skipped).
 
         # CLOSURE: Extract closure values and names.
         # For recursive/mutually recursive local functions, the closure may contain
--- a/tests/test_serialization_edge_cases.py
+++ b/tests/test_serialization_edge_cases.py
@@ -16,7 +16,31 @@ import torch
 
 sys.path.insert(0, "tests")
 
+from nnsight.intervention import serialization
 from nnsight.intervention.serialization import dumps, loads
+
+
+def _prohibited_error():
+    """``PicklingProhibitedError``, or skip if the lint check is not present.
+
+    ``PROHIBITED_MODULES`` / ``PROHIBITED_BUILTINS`` and their
+    ``PicklingProhibitedError`` were a lint-time check that raised before code
+    referencing ``os``, ``subprocess``, ``eval`` and friends was serialized for
+    NDIF. They are no longer in ``serialization.py``. Importing the exception at
+    call time -- and skipping rather than erroring -- keeps the rest of this
+    module collectable and makes these seven tests come back on their own if
+    the check is restored, instead of failing as ImportErrors that nothing runs.
+    """
+
+    error = getattr(serialization, "PicklingProhibitedError", None)
+
+    if error is None:
+        pytest.skip(
+            "serialization.PicklingProhibitedError is not present; the "
+            "lint-time prohibited-module check is not currently implemented"
+        )
+
+    return error
 
 
 def test_deserialized_helper_trace_uses_registered_linecache(tiny_model, tmp_path):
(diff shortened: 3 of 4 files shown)


JadenFiotto-Kaufman:
Hey @Hotragn ! Thank you so much for the recent PRs!

Right now im actively developing off the 0.8 branch here: https://github.com/ndif-team/nnsight/tree/0.8
Could you base your work off this instead? I'll still respond in your PRs as they are for now.

Hotragn:
Thanks — moved over to `0.8`, and closing this one: almost everything it fixes is already handled there.

Checked each piece against `0.8`:

- **Stale imports** — gone. `test_whitelist_serialization.py` and `test_serialization_edge_cases.py` are consolidated into `tests/test_serialization.py`, with no `SERVER_MODULES_WHITELIST` / `PicklingProhibitedError` references left.
- **The misleading `# SECURITY CHECK` comment and dead `captured_globals`** — gone from `0.8`'s `serialization.py`.
- **The CI gap** — gone. `0.8` runs `pytest tests/ -q --ignore=tests/vllm --ignore=tests/tp`, so the whole directory is covered rather than an enumerated list. That's a better fix than the one I proposed.
- **The order-dependent registration tests** — no longer present in that form.

So the whole PR is redundant on `0.8`. Nothing for you to action here.

Two things did come out of running the `0.8` suite, though:

1. **#698** — `tests/test_serialization.py::TestScopeFiltering` has two real failures on Python 3.14 (which `pyproject.toml` lists as supported). `_function_referenced_names` picks the function's `symtable` child by counting children, and PEP 649 adds an `__annotate__` table to every `def`, so the guard trips and the scope filter is inert — reinstating the closed-file capture its docstring describes. CI runs 3.12, where PEP 649 isn't active, so it doesn't show.
2. **#699** — the cross-invoke `.backward()` diagnostic from #694, which still reproduces on `0.8`, plus the batched-invoke gradient coverage `0.8` is missing.

Also updated #695 — most of what I raised there is resolved on `0.8`; only the policy question remains, and it's a much smaller one now.

Hotragn:
Will do — everything is on `0.8` now. Thanks for the pointer; it changed the picture quite a bit, so here's where things stand.

I re-checked all four PRs against `0.8` rather than porting them blind, and three of them turned out to be already fixed there:

- **#692** (`util.apply` skipping container subclasses) — `0.8`'s `apply` is a full rewrite that handles all of it, and `applyn` is gone. Verified `ModelOutput` narrowing works on `0.8` (`rel_err 2.1e-07`, distinct objects per invoke). Closed.
- **#693** (the batched-gradient assertion) — that test doesn't exist on `0.8`. Closed.
- **#696** (serialization tests + CI) — stale imports gone, misleading comment gone, and `pytest tests/` covers everything. Closed.

Two are still live, both verified against `0.8` first:

- **#698** — `TestScopeFiltering` has two genuine failures on Python 3.14. `_function_referenced_names` finds the function's `symtable` child by counting children, and PEP 649 gives every `def` an extra `__annotate__` table, so the guard trips and the scope filter silently falls back to the block rule — bringing back the closed-file capture its docstring warns about. `pyproject.toml` lists 3.14 as supported; CI runs 3.12, where PEP 649 isn't active, so it doesn't surface. One-line fix plus regression tests. Full suite goes to 856 passed / 0 failed.
- **#699** — the cross-invoke `.backward()` diagnostic (still reproduces on `0.8`), plus the batched-invoke gradient coverage the suite is missing: `test_backward.py` is single-invoke only and `test_batching.py` is all `@torch.no_grad()`, so nothing asserted an invoke's gradient is its own rows. Built on the existing `MLP` + `_BatchEnvoy` pattern, so it's exact and adds no model download.

#699 will show the two `TestScopeFiltering` failures until #698 lands, since it's based on clean `0.8`. Happy to stack it on #698 instead if you'd rather.

One note in case it's useful: `pyproject.toml` advertises 3.13 and 3.14 but CI only runs 3.12. #698 is exactly the kind of thing that gap hides — PEP 695 generics would trip the same `symtable` code path on 3.13. Glad to open a PR adding 3.13/3.14 to the test matrix if you want the coverage.

PR #699, opened 2026-08-24
feat(backward): explain the freed autograd graph in terms of invokes, and cover batched-invoke gradients

Hotragn:
Rebased onto `0.8` as requested in #696 — this supersedes #694, and folds in the salvageable half of #693. Verified the problem still reproduces on `0.8` before porting.

## Summary

Calling `.backward()` in more than one invoke of a trace fails inside `torch.autograd`:

```python
with model.trace() as tracer:
    with tracer.invoke(prompt_a):
        a1 = model.transformer.h[5].output
        with model.output.logits.sum().backward():
            grad_a = a1.grad.save()

    with tracer.invoke(prompt_b):
        a1 = model.transformer.h[5].output
        with model.output.logits.sum().backward():   # RuntimeError
            grad_b = a1.grad.save()
```

```
RuntimeError: Trying to backward through the graph a second time (or directly
access saved tensors after they have already been freed). ...
```

Accurate, but it points away from the cause. Each invoke calls `.backward()` exactly **once**, so "a second time" reads as describing someone else's code. And the real reason is a fact about nnsight rather than torch: **invokes are not separate runs.** Every invoke's input is combined into one batch and the model is called once, so the whole trace has a single autograd graph, and the first `.backward()` frees it for everything after it.

`docs/gotchas/backward.md` does cover `retain_graph`, but as *"if you call `.backward()` more than once on overlapping graphs"*. One backward per invoke does not read as that.

## Changes

**`_explain_freed_graph`** translates that one error — matching autograd's message (a bare `RuntimeError`, no type or code to key off, so string matching is the only option and it is kept narrow), stating the cause in nnsight's terms, and showing the fix while appending torch's original text:

```
RuntimeError: The autograd graph for this trace has already been freed.

Every invoke in a trace contributes to a single batched forward pass, so all
invokes share one autograd graph. An earlier `.backward()` in the same trace
freed the graph that this one needs.

Pass `retain_graph=True` to every `.backward()` except the last one:

    with model.trace() as tracer:
        with tracer.invoke(prompt_a):
            ...
            with loss.backward(retain_graph=True):
                grad_a = a1.grad.save()
        with tracer.invoke(prompt_b):
            ...
            with loss.backward():
                grad_b = a1.grad.save()

The same applies to two `.backward()` calls inside one invoke. If you do not
need the gradients together, run each backward pass in its own trace instead.

Original error from torch.autograd: Trying to backward through the graph a
second time ...
```

When the *failing* call already passed `retain_graph=True`, it says the earlier one did not, instead of suggesting a flag that is already set. Every other `RuntimeError` is re-raised untouched.

**Batched-invoke gradient coverage**, which the suite was missing entirely. `test_backward.py` only exercised single-invoke gradients, and `test_batching.py` is entirely `@torch.no_grad()` — so nothing asserted that an invoke's gradient is *its own rows*. `TestBackwardAcrossInvokes` pins that against plain-autograd references:

- two invokes, each reading its own gradient
- three invokes, parametrised over which one takes the gradient (first / middle / last), so each batch offset is exercised rather than only the final one
- a negative control: the result must not match another invoke's gradient
- edit isolation: `a.grad = a.grad * 3` in one invoke leaves the other untouched

Invokes get **different row counts** (2 / 3 / 1), so a slice read from the wrong offset shows up as a shape mismatch even where the values would be close. Built on the existing `MLP` + `_BatchEnvoy` pattern, which keeps the assertions exact — `allclose(atol=1e-6)` — and adds no model download; the whole file runs in ~16s.

## Behaviour is unchanged

`retain_graph=True` across invokes already worked, and each invoke's gradient was already correct — verified before writing the diagnostic. This makes the failure legible and locks the correctness in.

## Not done here

nnsight could retain the graph implicitly for all but the last invoke and make this disappear. That trades peak memory for convenience on every batched trace and changes a default, so it seemed like your call rather than something to fold into an error message. Happy to do it if you'd prefer.

## Verification

Without the diagnostic, 2 of the 15 tests in `test_backward.py` fail; with it, 15 pass. Full CPU suite on this branch: **857 passed, 7 skipped, 2 failed** — the two failures are `test_serialization.py::TestScopeFiltering` on Python 3.14, which predate this branch and are fixed in #698.

<sub>`0.8` @ 1f974f0, Python 3.14.3, `transformers` 5.15.1, `torch` 2.9.1, CPU.</sub>


Hotragn:
Rebased onto current `0.8` now that #698 is in, so the caveat in the description is resolved — the full CPU suite is green on this branch:

```
854 passed, 7 skipped, 1 xfailed, 27 warnings
```

No other changes; same single commit, just replayed onto `8f7546c` with no conflicts.

Hotragn:
Heads-up on the red CI here — it isn't this branch.

All 6 failures are in `tests/test_tensor_parallel_rules.py`, identical on both of my open PRs, neither of which touches TP. `transformers` 5.16 moved tensor parallelism onto DTensor and deleted the `all_gather` / `split` that `modeling/tp/fragments.py` imports. Bisected:

| transformers | `tests/test_tensor_parallel_rules.py` |
|---|---|
| 5.15.1 | 87 passed |
| 5.16.1 | 6 failed, 81 passed |

Filed as #702 with the full diagnosis — it's broader than the 6 tests, since the missing imports are function-local and only bite at runtime on hardware.

Everything else in the run is green (882 passed), and the local CPU suite on this branch passes with 5.15.1. Nothing to change here; happy to rebase once #702 is sorted.

Hotragn:

JadenFiotto-Kaufman:
Taking the docs half of this and dropping the diagnostic. Thank you for it. The
reasoning is worth writing down.

The diagnostic is both too broad and too specific. Too broad because it fires on
any freed-graph error in a trace, including a trace with no invokes at all. Two
`.backward()` calls in a plain `model.trace(prompt)` reach it, and the message
that comes back opens with "Every invoke in a trace contributes to a single
batched forward pass" and then shows twelve lines of `tracer.invoke(...)`, for
code that uses no invokes. The only line that covers them is the last one. That
is worse than torch's original error, which at least does not send you looking
for a batching concept you were not using.

Too specific because two backward passes really are two backward passes, and
`retain_graph` is a torch convention people are expected to know. Re-framing a
torch error as an nnsight one puts us on the hook for keeping the framing right
in every case it can be reached from, which is the trap above.

Something else worth knowing: `0.8` picked up the documentation independently in
the meantime. The TL;DR bullet and the `retain_graph=True` section both cover the
cross-invoke case with a worked example now. So what is left from this PR is the
two things that section did not say, and I have folded those into it rather than
adding a second section:

- an invoke's gradient is still its own rows
- separate traces are cheaper than retaining, when you do not need the gradients
  together

On the tests. They were the strongest part of the PR, and they went only because
they exist to pin the diagnostic. If you want to send the batched-gradient
coverage on its own, the parametrisation over which invoke takes the gradient and
the differing row counts so a wrong offset fails on shape, that stands by itself
and I would take it. Nothing currently asserts that a batched invoke's gradient
is its own rows.

PR #706, opened 2026-08-29
fix(tp): explain a transformers whose TP collectives are gone, and skip the rules it no longer describes

Hotragn:
Fixes the CI failures reported in #702. Based on `0.8`.

## Summary

transformers **5.16.0** rebuilt tensor parallelism on DTensor. The implementation moved to `transformers.distributed.tensor_parallel`, and `transformers.integrations.tensor_parallel` became a re-export shim — one that does **not** re-export `all_gather` and `split`, the two module-level helpers `_gather` and `_reshard` reassemble fragments with. They are gone, with no drop-in replacement (a DTensor is made whole through `redistribute` / `full_tensor`).

`pyproject.toml` pins `transformers` unbounded, so 5.16 is what CI installs today. **Every PR against `0.8` is currently red on the same six tests**, mine included:

```
6 failed, 879 passed, 8 skipped, 1 xfailed
```

The runtime breakage is wider than those six show. Both imports are function-local:

```python
def _gather(value, mesh):
    from transformers.integrations.tensor_parallel import all_gather
```

so nothing fails until a hook fires on the first sharded module — and then it's a bare `ImportError` naming a module that *does* still import, on a path that only runs on multiple GPUs.

## Changes

**1. One guarded import.** Both call sites now go through `_collectives()`, which turns a missing pair into `UnsupportedTransformersVersion` naming what changed and what to pin — the same shape as the `MINIMUM_TRANSFORMERS` refusal directly above it:

```
UnsupportedTransformersVersion: transformers 5.16.1 does not provide the
`all_gather` / `split` helpers that nnsight reassembles sharded activations
with, so a fragment cannot be made whole for an intervention to see. 5.16.0
rebuilt tensor parallelism on DTensor and removed them. Pin
`transformers<5.16`, or load this model on one GPU.
```

**2. The six tests skip where their subject doesn't exist.** Two compare `SIDES`/`UNSUPPORTED` against the upstream registry (5.16 registers `colwise_rep` and `rowwise_rep`, drops `embedding_colwise`); four monkeypatch `split`. All six assert against the `all_gather`/`split` implementation nnsight drives, so on a DTensor build they say nothing about whether nnsight is right.

The skip is keyed on **the helpers being importable**, not on a version number — so a transformers that still has them and adds a style fails these as loudly as before. Refusing to run isn't the same as refusing to notice.

## What I deliberately did *not* do

- **No upper version bound.** `test_the_floor_and_above_are_allowed` admits `6.0.0` on purpose. Making future versions refuse by default reverses that policy, and that's your call, not a side effect of a CI fix. I built one first and threw it away for this reason.
- **No new `SIDES` entries.** The table's own note says to read a style's `_prepare_input_fn`/`_prepare_output_fn` before adding one, *"the name alone misled on `moe_tp_experts`"*, and the starred rows were measured on Llama-3.2-3B at tp=4. `colwise_rep`/`rowwise_rep` belong to an implementation this version can't drive; entering them against it would put unmeasured rows into a table whose whole point is distinguishing measured from inferred. I have no multi-GPU box to measure on.

Porting `_gather`/`_reshard` to DTensor and re-measuring `SIDES` is the real follow-up — that's #702, and I'm happy to do the mechanical half if you can confirm the handoff points.

## Verification

Both sides of the break, full CPU suite:

| transformers | result |
|---|---|
| 5.15.1 | **860 passed, 7 skipped, 0 failed** |
| 5.16.1 | **854 passed, 13 skipped, 0 failed** |

The added test (`test_a_transformers_without_the_collectives_is_refused`) fails without the guard.

<sub>`0.8` @ 5be282b, Python 3.14.3, `torch` 2.9.1, CPU.</sub>

### DIFF
--- a/src/nnsight/modeling/tp/fragments.py
+++ b/src/nnsight/modeling/tp/fragments.py
@@ -170,9 +170,40 @@ def _shardable(tensor: torch.Tensor, world_size: int) -> bool:
     return tensor.is_floating_point() and tensor.dim() >= 1 and tensor.shape[-1] % world_size == 0
 
 
+def _collectives():
+    """transformers' ``(all_gather, split)``, the pair fragments are moved with.
+
+    Imported through here rather than at each call site so a transformers that
+    does not have them is one error, raised where it can be explained.
+
+    5.16.0 rebuilt tensor parallelism on DTensor: the implementation moved to
+    ``transformers.distributed.tensor_parallel``, and these two module-level
+    helpers are not part of what the old path re-exports -- they are gone, with
+    no drop-in replacement (a DTensor is made whole through
+    ``redistribute``/``full_tensor``). Left alone, the failure is an ImportError
+    from inside a hook on the first sharded module, naming a module that does
+    still import, on a code path that only runs on multiple GPUs.
+    """
+    try:
+        from transformers.integrations.tensor_parallel import all_gather, split
+    except ImportError as error:
+        import transformers
+
+        raise UnsupportedTransformersVersion(
+            f"transformers {transformers.__version__} does not provide the "
+            "`all_gather` / `split` helpers that nnsight reassembles sharded "
+            "activations with, so a fragment cannot be made whole for an "
+            "intervention to see. 5.16.0 rebuilt tensor parallelism on DTensor "
+            "and removed them. Pin `transformers<5.16`, or load this model on "
+            "one GPU."
+        ) from error
+
+    return all_gather, split
+
+
 def _gather(value: Any, mesh: Any) -> Any:
     """Every rank's slice of ``value``, concatenated back into the whole tensor."""
-    from transformers.integrations.tensor_parallel import all_gather
+    all_gather, _ = _collectives()
 
     return apply(
         value,
@@ -204,7 +235,7 @@ def _reshard(value: Any, mesh: Any) -> Any:
     are autograd functions. Guarded by the same predicate as the gather: a tensor
     that was never a fragment must not be cut down here.
     """
-    from transformers.integrations.tensor_parallel import split
+    _, split = _collectives()
 
     world_size = mesh.size()
     return apply(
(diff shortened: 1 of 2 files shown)


Hotragn:
Rebased onto `0.8` now that #707 is in, so this runs under the new matrix (3.12 and 3.14).

Worth flagging: with push CI now live on `0.8`, the branch itself is red — [run 33413414159](https://github.com/ndif-team/nnsight/actions/runs/33413414159) — and both jobs fail on exactly the six tests this PR fixes:

```
test (3.12): FAILED tests/test_tensor_parallel_rules.py ... (6)
test (3.14): FAILED tests/test_tensor_parallel_rules.py ... (6)
```

Identical on both, so nothing there is Python-version-specific — it is all transformers 5.16 (#702).

Local check against 5.16.1, which is what CI installs: `82 passed, 6 skipped, 0 failed`.

JadenFiotto-Kaufman:
Superseded by #709, which fixes this at the root. Thank you for it. Closing
unmerged rather than because it was wrong.

Two things it got right that I kept: keying the skip on capability rather than a
version number, and declining to add `SIDES` rows you could not measure. On the
second you were right to be cautious, and it turned out to be resolvable.
`colwise_rep` and `rowwise_rep` are constructed identically to
`colwise_gather_output` and `rowwise_split_input`, and both are now covered by a
tp=4 run on real hardware.

The reason I did not take the guard is that it sits on a path 5.16 never reaches.
Because the plan moved off the modules, see #702, `instrument` returns early for
every module and `_gather`/`_reshard` are never called. So `_collectives()` and
its `UnsupportedTransformersVersion` cannot fire, and the real failure is a
`mixed torch.Tensor and DTensor` error earlier, from a different cause. You could
not have seen that without a multi-GPU box. As it turns out transformers TP runs
on CPU over gloo, which is how I confirmed it, and that is now a CI test.

The skip-on-missing-helpers approach would also have turned CI green while TP
stayed entirely non-functional on 5.16. The red was the only signal that anything
was wrong.

PR #708, opened 2026-09-02
Remote Job Metadata

AdamBelfki3:
Adds `meta_data` to `ResponseModel`: what a remote job cost the server, reported
on the response that ends it.

```python
with model.trace(prompt, remote=True) as tracer:
    out = model.lm_head.output.save()

tracer.backend.meta_data.runtime                  # 0.42 seconds
tracer.backend.meta_data.max_mem_pct_by_gpu       # {'0': 20.0}
```

The tracer holds the backend, so no new API is needed on the common path. Read
it after the block exits — the backend runs in `__exit__`.

## What's here

- `MetaData` model in `schema/response.py`: `runtime`, `max_memory_usage`,
  `max_mem_by_gpu`, `max_mem_pct_by_gpu`, and `alloc_shortfall_by_gpu`.
- `RemoteBackend.meta_data`, populated in `note()` — the per-frame handling the
  blocking trace, the non-blocking poll and `await backend` all share.
  `stream()` bypasses `note()` by design, so it records on its own path.
- Recorded **before** `note()` raises, so a failed job's report survives and can
  be read from `except RemoteError:`. On an out-of-memory failure that report
  carries `alloc_shortfall_by_gpu`, which is the useful part.

## Testing

19 tests in `tests/test_remote_backend.py` covering both wire encodings (JSON
and `torch.save`), all four waiting modes, the failure path, unknown fields, and
the unreadable-report warning. Exercised end to end against a local NDIF stack
with a real GPU from the paired ndif branch.

### DIFF
--- a/docs/remote/ndif-overview.md
+++ b/docs/remote/ndif-overview.md
@@ -77,6 +77,27 @@ Status updates are `ResponseModel` objects carrying one of these (`src/nnsight/s
 
 The client renders these as a single in-place status line (animated spinner in terminals, an in-place HTML element in Jupyter). See `StatusDisplay` (`src/nnsight/intervention/backends/display.py:58`). There is no `STREAM` status — the old `tracer.local()` hybrid-streaming path does not exist here.
 
+## What the job cost
+
+The `COMPLETED` response also carries `meta_data` — what the run cost on the server. The backend keeps the last one it saw, and the tracer keeps the backend it ran on, so read it after the block exits (the backend runs in `__exit__`, so it is still `None` inside):
+
+```python
+with model.trace("Hello", remote=True) as tracer:
+    out = model.lm_head.output.save()
+
+print(tracer.backend.meta_data)
+# {'runtime': 0.42,                      # wall-clock seconds on the server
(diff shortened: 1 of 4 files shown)


koriyoshi2041 (commented):
There is a stale-report edge when a caller reuses an explicit blocking backend across traces. `Remotable.trace(..., backend=...)` accepts the backend as a public argument, but `note()` only assigns `self.meta_data` when the new response has a report.

The lifecycle reduces to:

```python
backend = RemoteBackend(MODEL_KEY, host="http://ndif.test")
backend.note(ResponseModel(id="first", status=Status.COMPLETED, meta_data=META))
backend.note(ResponseModel(id="second", status=Status.COMPLETED))  # older server / no report

assert backend.meta_data is None  # currently still META from "first"
```

That can make the second job appear to have the first job's runtime/GPU footprint. The existing intermediate-update test correctly requires `RUNNING` without metadata not to clear the current job's eventual report, so clearing in every `note()` would be too broad. Could the start of a new blocking request reset `meta_data`, with a two-request regression that preserves the intermediate-frame behavior?

PR #713, opened 2026-09-04
Preserve model forward signatures through Accelerate wrappers

### LINKED ISSUE #711: [Bug]: With multiple-gpus tracer fails with Qwen3.5
```python
import numpy as np
import torch # 2.12.1
from PIL import Image
from transformers import AutoModelForImageTextToText, AutoProcessor # 5.16.1

def make_inputs(processor, n_frames=4, size=(320, 224)):
    rng = np.random.default_rng(0)
    frames = [
        Image.fromarray(rng.integers(0, 255, size=(size[1], size[0], 3), dtype=np.uint8))
        for _ in range(n_frames)
    ]
    messages = [{"role": "user", "content": [
        *[{"type": "image"} for _ in frames],
        {"type": "text", "text": "Describe these frames."},
    ]}]
    text = processor.apply_chat_template(messages, add_generation_prompt=True, tokenize=False)
    return processor(text=[text], images=frames, return_tensors="pt")
from nnsight.modeling.vlm import VisionLanguageModel
REPO = "Qwen/Qwen3.5-0.8B"

nn_model = VisionLanguageModel(REPO, dtype=torch.bfloat16, device_map="auto", dispatch=True)
inputs2 = make_inputs(nn_model.processor)

with nn_model.generate(dict(inputs2), max_new_tokens=8, do_sample=False) as tracer:
    result = tracer.result.save()
```
With multiple GPUs this causes an error:
```
  with nn_model.generate(dict(inputs2), max_new_tokens=8, do_sample=False) as tracer:
  File "/projectnb/ivc-ml/env/cogmap/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
    return func(*args, **kwargs)
  File "/projectnb/ivc-ml/env/cogmap/lib/python3.12/site-packages/transformers/generation/utils.py", line 2483, in generate
    self._validate_model_kwargs(model_kwargs.copy())
  File "/projectnb/ivc-ml/env/cogmap/lib/python3.12/site-packages/transformers/generation/utils.py", line 1664, in _validate_model_kwargs
    raise ValueError(

ValueError: The following `model_kwargs` are not used by the model: ['mm_token_type_ids', 'pixel_values', 'image_grid_thw'] (note: typos in the generate arguments will also show up in this list)

```

koriyoshi2041:
Fixes #711.

When Accelerate replaces `module.forward` with `partial(new_forward, module)`, `Interleaver.wrap_module` currently uses the generic wrapper for both execution and introspection. Transformers generation validation then sees only `*args, **kwargs` and rejects valid multimodal inputs such as `pixel_values` and `image_grid_thw`.

This keeps execution routed through Accelerate while exposing the bound `_old_forward` signature on the NNsight wrapper. The regression uses a synthetic module and calls Transformers `_validate_model_kwargs`, so it runs on CPU without model weights or multiple GPUs.

Validation:

- focused regression: 1 passed
- pre-patch-equivalent signature rejects `pixel_values` and `image_grid_thw`
- patched signature accepts both and preserves execution
- `git diff --check`

### DIFF
--- a/src/nnsight/intervention/interleaver.py
+++ b/src/nnsight/intervention/interleaver.py
@@ -560,6 +560,11 @@ class Interleaver:
                 # e.g. accelerate's partial(new_forward, module) for device_map —
                 # unwrap to avoid reference cycle through partial.args
                 original_forward = instance_forward.func
+                introspection_signature = (
+                    inspect.signature(module._old_forward)
+                    if hasattr(module, "_old_forward")
+                    else inspect.signature(instance_forward)
+                )
             else:
                 original_forward = instance_forward
 
@@ -597,6 +602,9 @@ class Interleaver:
                 if source_accessor is not None:
                     return source_accessor(m, *args, **kwargs)
                 return m.__nnsight_forward__(m, *args, **kwargs)
+
+            if isinstance(instance_forward, partial):
+                nnsight_forward.__signature__ = introspection_signature
 
             module.forward = nnsight_forward
 
--- a/tests/test_interleaver_signature.py
+++ b/tests/test_interleaver_signature.py
@@ -1,0 +1,42 @@
+import inspect
+from functools import partial
+from types import SimpleNamespace
+
+import torch
+from transformers import GenerationMixin
+
+from nnsight.intervention.interleaver import Interleaver
+
+
+class MultimodalModule(torch.nn.Module, GenerationMixin):
+    config = SimpleNamespace(is_encoder_decoder=False)
+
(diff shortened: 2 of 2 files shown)


koriyoshi2041:
Closing this because the 0.8 pipeline rewrite superseded the old `wrap_module` path. Current `Controller` preserves the class-level forward signature, and the reporter confirmed they were on 0.7; the exact multi-device script is already fixed on 0.8. Rebasing this patch would restore code the new architecture intentionally removed.

PR #729, opened 2026-09-17
ndif: cache the remote environment per host

JadenFiotto-Kaufman:
## Why

`nnsight.ndif.get_remote_env()` fetches a server's `/env` (Python version plus `{import_name: version}`) and held it in one module-level global. It took no host, so:

- a caller using a per-call host (`remote="http://host:port"`, which `RemoteBackend` supports) read the **configured** server's environment, not the one its job runs on;
- a process that talks to two servers got whichever was asked first.

The case that prompted this: a downstream engine (causalab running on NDIF) checks, before it queues a job, that the server has the same `causalab` and `nnterp` versions as the client, against whichever host that job targets. With one global, that check answers for the wrong server.

## What changes

```python
ndif.resolve_host(host=None) -> str
ndif.get_remote_env(host=None, *, force_refresh=False) -> dict
ndif.set_remote_env(env, host=None) -> None
ndif.clear_remote_env(host=None) -> None      # None clears every host
nnsight.compare(host=None) -> EnvComparison
```

- **Per-host cache.** `get_remote_env` resolves `host` (the configured `CONFIG.API.HOST` when `None`, read at call time), fetches once per host, and caches under the resolved URL. `force_refresh=True` re-fetches that host only. `compare(host)` reads the same cache.
- **One spelling per server.** `resolve_host` drops a trailing slash and checks the scheme. `RemoteBackend.__init__` resolves its host through it (it held the scheme check before), so `remote="http://h:5001/"` and `get_remote_env("http://h:5001")` share a key. A trailing slash on a per-call host also stops producing `//request` URLs.
- **Errors.** A failed fetch caches nothing, so the next call tries again, and a failed `force_refresh` leaves the cached environment in place. The error is a `RuntimeError` naming the host; a server with no `/env` route gets its own message.
- **A seam for test harnesses.** `set_remote_env` / `clear_remote_env` seed and empty the cache without a network call. The backing dict is `ndif._REMOTE_ENVS`; the old `ndif._REMOTE_ENV` global is gone rather than kept as a second source of truth, so a harness that `monkeypatch.setattr`s it fails loudly and moves to `set_remote_env`.
- **Compatibility.** Every existing call keeps working: `get_remote_env()`, `get_remote_env(force_refresh=True)` (used by ndif's live test suite), the positional `get_remote_env(True)`, and `compare()`.

`RemoteBackend` does not fetch the remote environment on submit, before or after this PR: `_serialize` calls `pull_env()`, which only scans the local environment and registers `"local"` modules for by-value pickling.

## Tests

`tests/test_ndif.py::TestRemoteEnv` (15 tests, no network; the module's HTTP getter is replaced by a counting fake): one fetch per host, two hosts cached independently, `force_refresh` re-fetches one host, default-host resolution at call time, trailing-slash normalization (including a slash on `CONFIG.API.HOST`), `RemoteBackend.host` and the cache key agree, seeding and clearing through the public helpers, a failed fetch is not cached, a failed refresh keeps the cached env, a 404 on `/env` names the host, the request URL is `{host}/env`, and `compare(host=...)` uses that host's env.

Ran `tests/test_ndif.py tests/test_config.py tests/test_remote_backend.py tests/test_login.py tests/test_modeling.py tests/test_serialization.py tests/test_deprecations.py`: 232 passed.

## Docs

`docs/remote/env-comparison.md` (comparing against a specific server, per-host caching, seeding the cache), `docs/reference/api-quick-reference.md`, and one line in `docs/reference/version-history.md`.


koriyoshi2041 (approved):
Reviewed head 524c33fcc425e0bde908189093000fd17ce9f458. The cache key, RemoteBackend target, and request URL now share resolve_host normalization; default-host lookup remains call-time dynamic; failed initial fetches cache nothing while failed refreshes preserve the previous value because assignment occurs only after a successful fetch. The focused tests cover independent hosts, trailing-slash aliases, refresh isolation, runtime config changes, seeding/clearing, failure retry, 404 diagnostics, request routing, and compare(host). I found no blocking correctness issue.

